PR #12, opened 2014-12-05
When used with enhanced-require on node.js it throws an error because emitFile is not existent.

### LINKED ISSUE #2: emitFile is required from module system
Curious why an error is thrown for this. Should it not fail silently (or still warn), but at least return the same path to the file and just not call `emitFile`?

I noticed the issue raised over at `enhanced-require`:  https://github.com/webpack/enhanced-require/issues/3, and I understand why, but here's my use case:

I'm using webpack via the Node.js API to server-side render a site built heavily with [React](facebook.github.io/react). In order to benefit from the server-side rendering, the client side code must render the same thing when the React component is mounted to the DOM. Using `require('url!mylogo.png');` renders the logo fine but where `emitFile` can't be used I need to at least get the same path to the image for the React client side rendering to match the prerender from the server.

In order to get this to work properly I amended the loader as follows:
-   Remove error thrown when no emitFile present (ie: via node api / enhanced-require)
-   Check for emitFile and don't call it if it's not there

I also had to set `__webpack_public_path__` at runtime if it was not set to avoid an issue there.

Then it works. Hopefully you understand the React bit. I can see why you might want to alert the user but for my purposes it helps if the loader fails silently when emitFile is not present.

bitplanets:
Fix #2

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,14 +7,15 @@ var path = require("path");
 
 module.exports = function(content) {
 	this.cacheable && this.cacheable();
-	if(!this.emitFile) throw new Error("emitFile is required from module system");
 	var query = loaderUtils.parseQuery(this.query);
 	var url = loaderUtils.interpolateName(this, query.name || "[hash].[ext]", {
 		context: query.context || this.options.context,
 		content: content,
 		regExp: query.regExp
 	});
-	this.emitFile(url, content);
+    // When used with enhanced-require on node.js it throws an error because
+    // emitFile is not existent.
+	this.emitFile && this.emitFile(url, content);
 	return "module.exports = __webpack_public_path__ + " + JSON.stringify(url);
 }
 module.exports.raw = true;


bitplanets:
Any news about this?

jhnns:
> Just ignoring isn't valid.

I have to admit that it *could* be valid but I'm not sure about this.

Usually you don't need the file-loader in combination with enhanced-require. But I understand that there are situations where the file-loader is not necessary but in use. For example when the code is isomorphic (running in the browser and in node).

What does the file-loader return then? The original path to the file?

However, these situations can be dangerous because the developer might not be aware that the file-loader returns stuff that is not meant to be sent to the client (such as internal paths).

bitplanets:
>What does the file-loader return then? The original path to the file?

Returns the same as it would return if it were on the client side. Is this possible?

sokra:
https://github.com/webpack/file-loader/issues/2#issuecomment-40043276

bitplanets:
Should send the same URL as it would send on the client side. Isn't that
possible?

jhnns:
I think we should continue this discussion at #2

sokra:
fixed by other PR

PR #16, opened 2015-03-06
added support for source maps

thetalecrafter:
If the webpack configuration is set to generate source maps, and a source map for the file comes into the loader, emit the map in addition to the file.

This is particularly useful for css preprocessing to a file:

```
loaders: [ 'file?name=[name].css', 'cssnext' ]
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@
 var loaderUtils = require("loader-utils");
 var path = require("path");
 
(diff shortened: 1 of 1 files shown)


thetalecrafter:
hmm, need to add the comment.

thetalecrafter:
I will redo this.

PR #33, opened 2015-08-18
Add "noEmit" option to don't copy files for server-side builds

jeron-diovis:
Related to #2. 
Not sure I understand problems with enhanced-require and __webpack_public_path__. 
For me everything works, the only problem is that for server-side build we don't actually need to emit files to output dir (which can actually differ from client's one), we only need a url for resource.
So just need ability to skip emitting, nothing more.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,9 @@ module.exports = function(content) {
 		content: content,
 		regExp: query.regExp
 	});
-	this.emitFile(url, content);
+	if (!query.noEmit) {
+		this.emitFile(url, content);
+	}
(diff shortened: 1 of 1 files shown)


3went2:
+1

aunz:
+1

hagaiak:
If anyone's searching for a quick solution, there are these 3 loaders:
[fake-file-loader](https://github.com/dferber90/fake-file-loader)
[fake-url-loader](https://github.com/dferber90/fake-url-loader)

Which work exactly as the original loaders, except that they do not actually emit any file.

tabazevedo:
+1

sokra:
fixed by other PR

PR #41, opened 2015-10-25
Remove unused dependency

kossnocorp:
Remove `var path = require("path");` because it's not used anywhere in the source.


jordansexton:
This was fixed by c5959984d13fdc27ce9d30cf429755bc89149abe from #25 and can be closed.

PR #42, opened 2015-10-25
Added a "publicPath" config option

jordansexton:
For [domain sharding](http://abhishek-tiwari.com/post/CloudFront-design-patterns-and-best-practices) with hashed filenames, or splitting up build output assets by type or based on their source directories in a project, it's helpful to be able to differentiate the public path of the files emitted by the loader.

This PR is for a patch I've been using for a while now, and is a non-breaking change. I can also add tests for it if you like. It allows for a new query option, `path`, to be provided. `query.path` can be either a:

- String (e.g. `https://static.cdn.net/`)

  Since static assets are often served from a separate source than the application, this may be different from `options.output.publicPath`, and will be prepended to the output file path by the loader.

- Function, taking a single `path` argument, the file path from the loader, e.g.:

```javascript
function shard (path) {
    var hash = 0;
    var digit;
    for (var i = 0, length = path.length; i < length; i++) {
        hash = (((hash << 5) - hash) + path.charCodeAt(i)) & 0xFFFFFFFF;
    }
    digit = Math.abs(hash % 10);
    return 'https://static' + digit + '.cdn.net/' + path;
};
```

This allows assets to be sharded across domains based on a deterministic filename hash. Having deterministic output filenames is important for caching and linking of the assets (using React or another isomorphic rendering mechanism).

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,12 +8,22 @@ module.exports = function(content) {
 	this.cacheable && this.cacheable();
 	if(!this.emitFile) throw new Error("emitFile is required from module system");
 	var query = loaderUtils.parseQuery(this.query);
+	var configKey = query.config || "fileLoader";
+	var config = this.options[configKey] || {};
 	var url = loaderUtils.interpolateName(this, query.name || "[hash].[ext]", {
 		context: query.context || this.options.context,
 		content: content,
 		regExp: query.regExp
 	});
+	var publicPath = config.publicPath;
+	if (publicPath) {
+		publicPath = (typeof publicPath === "function") ? publicPath(url) : publicPath + url;
+		publicPath = JSON.stringify(publicPath);
+	}
+	else {
+		publicPath = "__webpack_public_path__ + " + JSON.stringify(url);
+	}
 	this.emitFile(url, content);
-	return "module.exports = __webpack_public_path__ + " + JSON.stringify(url) + ";";
+	return "module.exports = " + publicPath + ";";
 }
 module.exports.raw = true;


index.js
```diff
@@ -14,7 +14,15 @@ module.exports = function(content) {
 		content: content,
 		regExp: query.regExp
 	});
+	var path = query.path;
```
hamczu: I would use different name to avoid confusion with "[path]" placeholder - e.g. "base" or "publicPath"

hamczu:
This is really important and needed change so I would suggest to create proper test cases. 

btw. funny enough I had to write my own implementation because I hadn't discovered this PR before.

jordansexton:
@hamczu Would you be able to provide your implementation? I'm going to revise this PR a bit and want to see your ideas first.

jordansexton:
I've updated this PR per [webpack's documentation](https://webpack.github.io/docs/how-to-write-a-loader.html#programmable-objects-as-query-option) for loader's with programmable config, as this option uses a function to dynamically generate the public path.

I'll follow up soon with tests.

index.js
```diff
@@ -8,12 +8,22 @@ module.exports = function(content) {
 	this.cacheable && this.cacheable();
 	if(!this.emitFile) throw new Error("emitFile is required from module system");
 	var query = loaderUtils.parseQuery(this.query);
+	var configKey = query.config || "fileLoader";
+	var config = this.options[configKey] || {};
```
hamczu: The only thing I question here, why publicPath option cannot be defined in the query? I think this is a webpack convention. 

By doing this, one could define publicPath option as a string in different loader, which is very convenient.

hamczu:
@jordansexton hey, I didn't my put my implementation here, because it was still "in progress". I think yours is ok, I just have one comment (please see the code) about the config format. btw. how do you achieve effect that images with url replaced to CDN are not copied into dist?

@sokra can we have that PR merged? Or, if there's another way to dynamically change publicPath for specific loader, please tell us how.

hamczu:
@jordansexton I created my own PR, because I really need that option now :)
https://github.com/webpack/file-loader/pull/57

Differences:
- support for configKey (fileLoader) and query in the same time
- tests
- (TBD) optional skipping file emit

index.js
jordansexton: `publicPath` cannot be defined in the query because the query is converted to and from JSON, so only JSON data types are supported (meaning the function gets cast to a string).

jordansexton:
I mentioned this in the line note above, but putting a function in the query options is not idiomatic for webpack, as the query (at least until webpack 2 comes out, I think) must be `JSON.stringify`-able.

I also have [my own emitFile option implementation](https://github.com/jordansexton/file-loader/commit/c33edc629ac406ab97f3f8ee88e3570b41cd1fae), and there's an open PR for one (#33), so it might be valuable to separate that feature into a separate PR since it's not related to the `publicPath` option.

jordansexton:
@hamczu 
> how do you achieve effect that images with url replaced to CDN are not copied into dist?

I don't quite understand what you mean, but I can illustrate my project structure and build process a little bit if it helps. The project looks something like this (simplified):

```
application/
client/
    images/
static/
    images/
public/
    application/
    images/
```

I have a gulp task that first preprocesses all images in `client/images`. It creates multiple sizes, optimizes them, etc. and outputs those modified copies into `static/images`.

Webpack runs on React-based JS code in `application` which uses `file-loader`, e.g.

```jsx
<img
    src={ require('../static/images/image.png') }
    srcSet={
        require('../static/images/large/image.png') + ' 1x, ' +
        require('../static/images/large2x/image.png') + ' 1.5x'
    }/>
```

The loader, when provided with a public path function, [generates a deterministic hash from the filename](https://gist.github.com/jordansexton/2d62644556c4e49da116).

Webpack outputs everything to `public`, the directory contents of which all get pushed to CloudFront when deployed. The output JS will render HTML:

```html
<img
    src="https://static0.website.com/images/image.415158a92b93.png"
    srcset="https://static0.website.com/images/large/image.415158a92b93.png 1x, https://static1.website.com/images/large2x/image.87de0768e265.png 1.5x"/>
```

hamczu:
That's a one long-going PR :) @jordansexton I meant when you do require your image i.e. require('../static/images/image.png'), then this file is being copied to /dist during the build.
That's why I had to implement a flag to not emit a file once it's hosted somewhere on CDN.

jordansexton:
Still not sure I understand.

You want it to be emitted to `dist`/`public` and named with a hash. Then you want to put it on the CDN. Why would you want to not emit the file? If you don't emit it, you don't get the hashed filename necessary for best long-term caching.

jordansexton:
And I don't mean to say there's no use for your option to override `emitFile`, just that it makes sense to me to split into two separate PRs since they're unrelated. That said, this project hasn't been updated since either of our requests :) Maybe @sokra would be willing to chime in?

sokra:
fixed by other PR

sokra:
sharding should be a separate PR.

PR #58, opened 2016-02-28
Update README.md

mrpoptart:
Add installation instructions. 

Many of the other webpack modules include installation instructions, 
this one should, too.


sokra:
Thanks

PR #64, opened 2016-03-30
Added support for not emitting files for prerendering script.

pauldotknopf:
The css-loader supports "css/locals". This will export the appropriate path, but will not emit any files. This file serves the same purpose. You can now use 'file/locals' to get a url reference, but without emitting a file. Files are typically emitted with a second webpack configuration that doesn't use "locals" anywhere.

### DIFF
--- a/locals.js
+++ b/locals.js
@@ -1,0 +1,17 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+var loaderUtils = require("loader-utils");
+
+module.exports = function(content) {
+	this.cacheable && this.cacheable();
+	var query = loaderUtils.parseQuery(this.query);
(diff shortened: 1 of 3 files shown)


theverything:
@pauldotknopf you'll need to add the `locals.js` file to the `files` property in `package.json`

pauldotknopf:
@theverything, added

pauldotknopf:
@theverything, Who is managing this repo? @sokra?

theverything:
@pauldotknopf I believe so. Looks like he pops in every once in awhile and merges in some pull requests. Hopefully yours will get merged in soon. In the meantime I just forked the repo and added your pull request. I'm just installing the package from my github fork for now.

elyobo:
Huh, only saw this after opening #74; I've taken a different approach though, just allowing the emission of files to be specified as part of the query instead, which involves less copy and paste.

sokra:
fixed by other PR

PR #67, opened 2016-04-18
Path Fix

jenius:
Weird edge case, but when there was a path from `dumpDirs` being added it had a leading slash where no other path did. This actually was messing up our static plugin in roots-mini.


jenius:
Haha oops, wrong repo!

PR #74, opened 2016-06-16
Make emitting a file optional.

### LINKED ISSUE #59: Ignore file output
Hi

I love this plugin but in a isomorphic app you don't want the server-side app to generate the images, but you still want the correct url of the resource.

I searched but I could not find any solution, so I forked this module and then just deleted the line that saves the file.

I published it as `ignore-file-loader`. https://github.com/kevinsimper/ignore-file-loader/

elyobo:
When building server side packages there may be no need to emit a file;
this makes file emission optional, while keeping the default behaviour
as is.

Resolves #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,15 @@ var url = require("file!./file.png");
 By default the filename of the resulting file is the MD5 hash of the file's contents 
 with the original extension of the required resource.
 
+By default a file is emitted, however this can be disabled if required (e.g. for server
+side packages).
+
+``` javascript
+var url = require("file?emitFile=false!./file.png");
+// => returns the public url but does NOT emit a file
+// => returns i. e. "/public-path/0dcbbaa701328a3c262cfd45869e351f.png"
+```
+
 ## Filename templates
 
 You can configure a custom filename template for your file using the query
--- a/index.js
+++ b/index.js
@@ -13,7 +13,10 @@ module.exports = function(content) {
 		content: content,
 		regExp: query.regExp
 	});
-	this.emitFile(url, content);
+
+	if (query.emitFile === undefined || query.emitFile) {
+		this.emitFile(url, content);
+	}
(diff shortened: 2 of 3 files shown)


elyobo:
Note: similar in intent to #64.

pauldotknopf:
@elyobo, what if I use the same source file for both server and client? When I use webpack for my client side application, will files still be emitted?

elyobo:
If you set it up correctly, yes.  I use different webpack configurations for my client and server builds, enbabling emission on the client build, but not on the server build.

pauldotknopf:
Oh, so you mean adding ```emitFile=false``` to the actual entry point file in the webpack config, and not the ```resolve``` calls?

If so, are you saying that ```emitFile``` is recursive?

elyobo:
You've lost me there :D

Server webpack loaders config includes this (using image-loader as well)
```
      {
        test: /\.(jpe?g|png|gif|svg)$/i,
        loaders: [
          'file?hash=sha512&digest=hex&name=img/[name].[hash].[ext]&emitFile=false',
          'image-webpack?bypassOnDebug&optimizationLevel=7&interlaced=false',
        ]
      },
```

Client webpack loaders config includes this
```
        {
          test: /\.(jpe?g|png|gif|svg)$/i,
          loaders: [
            'file?hash=sha512&digest=hex&name=img/[name].[hash].[ext]&',
            'image-webpack?bypassOnDebug&optimizationLevel=7&interlaced=false'
          ]
        },
```

And then now I can import files anywhere and just get the URL in the server builds, but get the file emitted and the URL in the client builds.

```
import myImage from './image.jpg';
```

pauldotknopf:
Ah, ya, sorry, brain fart.

sokra:
Thanks

digz6666:
Thanks!

PR #104, opened 2016-11-16
Add -loader suffix to README to reflect new loader syntax webpack 2 requirement

kevinzwhuang:
Summary:
----

Webpack v2.1.0-beta.26 introduced a breaking change to require loaders to have the `-loader` suffix. 

This PR updates the README to show that pattern as an example instead of using the shorthand naming.

cc: @sokra


SpaceK33z:
Thanks

PR #122, opened 2017-02-03
fix: allow the usage of an empty string as publicPath

mlegenhausen:
This pull request enforces a more strict check of `publicPath` to prevent the interpretation of empty string as false value. This allows to disable the prefixing of `publicPath` with `__webpack_public_path__` without the need to provide a function that simply returns the `url` as given.

Current work around would be to use

```js
query: {
  publicPath: (url) => url
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,7 @@ module.exports = function(content) {
 		 : config.outputPath + url
 	}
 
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Closed in favour of #145. Thx

Issue #126, opened 2017-02-22
fileLoader option in webpack config object for webpack 2 is not allowed


KingMario:
In webapck 2,  the default option `fileLoader` in webpack config object is not allowed. `should NOT have additional properties` exception will be thrown.

In webpack 1, it works:
```
webpackConfig = {
  // entry, output, module ...
  fileLoader: {
    publicPath: function (url) {
      return `{{ "${url.replace(new RegExp(`^${config.build.assetsSubDirectory}/assets/`), '')}" | asset_url }}`
    }
  }
}
```
Here the publicPath callback function is used for shopify theme backend template.

For webpack 2, one can use `loader` with publicPath callback function, but it may have potential conflict with other loaders.

I just pulled PR #127, which will allow dot-separated path such as `loader.fileLoader` as the key for additional options for file-loader. With the original default `fileLoader` remains, and the same options fetching with non-dot-separated path, the change will not break any existing configuration.

michael-ciniawsky:
Discussion moved to #127

PR #127, opened 2017-02-22
Use dot-separated path such as loader.fileLoader for additional options

KingMario:
Use dot-separated path such as loader.fileLoader for additional options for webpack 2

feature

no

no

**Summary**

As mentioned in Issue #126, when used in webpack 2, with default settings, no additional options is allowed to add to the root of the webpack config, that is, no fileLoader option is allowed for additional publicpath callback function.

There is a loader option for webpack 2 allows any object. We can set query.config to `loader`.
A better way to do so is set query.config to `loader.fileLoader` that will avoid conflict with other loaders.

The new getOptions function will be used here to get the webpack options with dot-separated path.

no

**Other information**
Not much information about `loader` option in webpack config is published, the purpose and usage of this property may change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,13 +4,29 @@
 */
 var loaderUtils = require("loader-utils");
 
+function getOptions (obj, key) {
+	if (!key) {
+		return {}
+	}
+	var keyPaths = key.split('.');
+	var Options = obj;
+	for (var path of keyPaths) {
+		if (Options) {
+			Options = Options[path];
+		} else {
+			return {};
+		}
+	}
+	return Options || {};
+}
+
 module.exports = function(content) {
 	this.cacheable && this.cacheable();
-	if(!this.emitFile) throw new Error("emitFile is required from module system");
+	if (!this.emitFile) throw new Error("emitFile is required from module system");
 
 	var query = loaderUtils.parseQuery(this.query);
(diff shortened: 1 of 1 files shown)


michael-ciniawsky:
Could you please elaborate on this futher? What could be potentionally conflicting with other loaders like you mentioned in #126 ? (example) :)

michael-ciniawsky:
Closing do due to inactivity, address the comments and feel free to reopen :)

KingMario:
For webpack 2,  if I want to add a publicPath function, I'll use the loader option just in the root of webpack config.
```
module.exports = {
  // other options
  loader: { // Expose custom values into the loader context.
    publicPath: function (url) {
      return `{{ "${url.replace(new RegExp(`^${config.build.assetsSubDirectory}/assets/`), '')}" | asset_url }}`
    }
  }
}
```

What if another loader need a different publicPath function? So I suggest to change it to:
```
module.exports = {
  // other options
  loader: {
    fileLoader: {
      publicPath: function (url) {
        return `{{ "${url.replace(new RegExp(`^${config.build.assetsSubDirectory}/assets/`), '')}" | asset_url }}`
      }
    }
  }
}
```

That will have no potential conflict with any other loaders.

PR #130, opened 2017-03-05
refactor: apply `webpack-defaults`

### LINKED ISSUE #148: webpack-defaults upgrade
Addition of `webpack-defaults` & associated refactoring as a part of the next `Major` release

Issue exists for status tracking across the organization. 

>Please do not close

bebraw:

Refactoring.

The old tests should work.

**Summary**

This PR integrates **webpack-defaults** to the project.

Yes. Will enforce NodeJS > 4.3 via engines.

Must release as semver `MAJOR`

Closes #148


src/index.js
```diff
@@ -0,0 +1,59 @@
...
+    publicPath = JSON.stringify(
+      typeof config.publicPath === 'function' ? config.publicPath(url) : config.publicPath + url,
+    );
+  }
+
+  if (query.emitFile === undefined || query.emitFile) { // eslint-disable-line no-undefined
+    this.emitFile(outputPath, content);
+  }
+
+  return `module.exports = ${publicPath};`;
```
michael-ciniawsky:
```js
// webpack 2
if (loader.runner === 2) return `export default ${publicPath}`

return `module.exports = ${publicPath}`
```

michael-ciniawsky: ?

package.json
```diff
@@ -4,20 +4,47 @@
...
+    "jest": "^19.0.2",
+    "lint-staged": "^3.3.1",
+    "nodemon": "^1.11.0",
+    "nsp": "^2.6.2",
+    "pre-commit": "^1.2.2",
     "standard-version": "^4.0.0",
-    "mocha": "~1.21.3"
+    "webpack-defaults": "^0.1.1"
   },
   "scripts": {
```
michael-ciniawsky: Please sort them 😛 e.g

```js
"lint": "eslint --cache src test",
"test": "jest",
"test:coverage": "jest --collectCoverageFrom='src/**/*.js' --coverage",
"travis:lint": "yarn run lint && yarn run nsp",
"travis:test": "npm run test",
"travis:coverage": "yarn run test:coverage"
```

test/optional-file-emission.test.js
```diff
@@ -1,29 +1,27 @@
-var should = require("should");
-var fileLoader = require("../");
+const fileLoader = require('../');
```
d3viant0ne:
```
import fileLoader from '../';
```

src/index.js
```diff
@@ -0,0 +1,59 @@
+/*
+  MIT License http://www.opensource.org/licenses/mit-license.php
+  Author Tobias Koppers @sokra
+*/
+const loaderUtils = require('loader-utils');
```
d3viant0ne: `import loaderUtils from 'loader-utils';`

src/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = function fileLoader(content) {
```
d3viant0ne: `export default function fileLoader(content) {`

src/index.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+  }
+
+  if (query.emitFile === undefined || query.emitFile) { // eslint-disable-line no-undefined
+    this.emitFile(outputPath, content);
+  }
+
+  return `module.exports = ${publicPath};`;
+};
+module.exports.raw = true;
```
d3viant0ne: `export const raw = true;`

src/index.js
bebraw: That's original code I moved over. Looks like it detects webpack version. This is actually something I can do in certain other place...

package.json
bebraw: Any thoughts on this @sapegin? Is this something we can fix at mrm?

test/correct-filename.test.js
```diff
@@ -1,73 +1,70 @@
-var should = require("should");
-var path = require("path");
-var fileLoader = require("../");
+const fileLoader = require('../');
```
d3viant0ne: `import fileLoader from '../';`

src/index.js
bebraw: Yeah, let's do the change you proposed in another PR. 👍

package.json
d3viant0ne: `scripts` aren't reordered automatically so should be easy enough, i'll PR it into defaults in a minute.

test/optional-file-emission.test.js
bebraw: Yeah, I'll port those. 👍

src/index.js
michael-ciniawsky: Yep agreed, needs to be tested proberly, it bited me yesterday in `postcss-loader`+ `style-loader || extract-text-plugin` where es2015 modules don't work atm

package.json
michael-ciniawsky: 👍

test/correct-filename.test.js
d3viant0ne: Also needs to be moved into the `src` folder. Tests should reside with the code they are testing ( or that is what was discussed ).

The `build`  task ignores them though we need to put thought into how we want to handle libs with extensive `fixtures`

bebraw: Ah, yeah. I forgot to do that. Good call.

bebraw: Maybe we end up with both. Unit tests below `src` and fixtures/integration tests below `test`.

d3viant0ne: Looking at some of the libs, maybe a `test` folder is going to be the cleaner way to do this. Jest is smart enough to handle the preprocessing regardless of where the tests reside.

src/index.js
```diff
@@ -0,0 +1,59 @@
...
+      typeof config.publicPath === 'function' ? config.publicPath(url) : config.publicPath + url,
+    );
+  }
+
+  if (query.emitFile === undefined || query.emitFile) { // eslint-disable-line no-undefined
+    this.emitFile(outputPath, content);
+  }
+
+  return `module.exports = ${publicPath};`;
+};
```
d3viant0ne: Shouldn't be a semi here

bebraw: I think that was forced by the linting rules.

d3viant0ne: I promise that will fail if you run `yarn run lint`

bebraw: It's actually the opposite then:

```
.../file-loader/src/index.js
  58:2  error  Missing semicolon  semi
```

It's because of `module.exports = function fileLoader(content) {...};`.

package.json
sapegin: Not sure would be the best approach here. Sorting everything (including existing scripts) feels too destructive. Also would be nice to do it not just for `scripts`. @bebraw Good feature for webpack-merge? ;-)

bebraw: I'm not sure. The language specification doesn't really say anything about key order if I remember right.

Let's skip this problem for now. Open an issue at webpack-defaults repo, though.

bebraw: Note that npm has a tendency to do its own sorting when it touches **package.json** somehow so I'm not sure if it's worthwhile to spend too much time on fixing this.

d3viant0ne:
I pulled down the branch and made the suggested changes so I could check the transpiled output.
You should probably double check it but imo it looks good.

bebraw:
@d3viant0ne Ok, feel free to force push your changes over the PR. 👍

d3viant0ne:
@bebraw @sapegin - This is ready for a final look. 

- I've made the updates discussed above.
- Updated & published `webpack-defaults@0.2.1`
- Removed all the Bithound webhooks in their application across the org.

bebraw:
The only part that's a little puzzling to me is the handling of imports. Example:

```
file-loader [feat-webpack-defaults] $ npm run build

> file-loader@0.10.1 prebuild .../file-loader
> yarn run clean:dist

yarn run v0.20.3
$ del-cli dist
✨  Done in 0.50s.

> file-loader@0.10.1 build .../file-loader
> cross-env NODE_ENV=production babel -s true src -d dist --ignore 'src/**/*.test.js'

src/index.js -> dist/index.js
file-loader [feat-webpack-defaults] $ node
> require('./')
.../file-loader/dist/index.js:5
import loaderUtils from 'loader-utils';
^^^^^^
SyntaxError: Unexpected token import
```

It might make sense to force preprocessing imports too and expose through `module.exports`. Basically Node's lack of ES6 module support bites right here.

This isn't an issue if you use the loader within webpack, but it will be problematic if you use **loader-runner** and such.

d3viant0ne:
Open an issue in defaults so we can figure out the correct output for "all the things"

bebraw:
Sure. 👍

bebraw:
https://github.com/webpack-contrib/webpack-defaults/issues/23 for further discussion.

.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "webpack",
+  "globals": {
```
sapegin: I think we don’t need this anymore, since webpack preset includes `jest` environment.

bebraw: Yeah, safe to drop. 👍

d3viant0ne:
@bebraw this should transpile down properly. Still needs to have the version bumped once the next patch of defaults is released but the actual change is applied in 35ad7fa

bebraw:
Yup. Basically

```
file-loader [feat-webpack-defaults] $ node
> require('./')
{ raw: true, default: [Function: fileLoader] }
```

now. This leaves one decision - how to handle `default`.

bebraw:
Ready for a review. I changed the module entry point as discussed at https://github.com/webpack-contrib/webpack-defaults/issues/23.

src/index.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+  }
+
+  if (query.emitFile === undefined || query.emitFile) { // eslint-disable-line no-undefined
+    this.emitFile(outputPath, content);
+  }
+
+  return `export default = ${publicPath};`;
+}
+export const raw = true;
```
d3viant0ne: One thing of note, we are going to need a consistent way to handle the proxy + libs. While it seems like overkill on a single file library like file-loader, applying the same to something like css-loader would be a bit messy.

Proposing we as a standard, leave `cjs.js` in `./src` and move any library files into `./src/lib`

bebraw: Another way, go with `_cjs.js` so at least it will stand out as glue code. No need for `src/lib` then.

d3viant0ne: Still doesn't solve the issue with the more complicated libs imo. For the sake of simplicity, particularly with defaults upgrade / update still think libs in `lib` would be cleaner. Not that I want to endlessly debate the topic.

Take a look at some of our more complicated loaders / plugins, just hammering everything into `./src` stands to get a little messy.

d3viant0ne:
@bebraw @sapegin @michael-ciniawsky  Last feature to merge prior to 1.0 hit npm this morning. I've rebased this and picked up the relative url feature. Fixed it's test and updated to the latest defaults. 

This needs a once over and then it's time to bring this one home.

package.json
```diff
@@ -4,20 +4,49 @@
   "author": "Tobias Koppers @sokra",
   "description": "file loader module for webpack",
   "files": [
-    "index.js"
+    "index.js",
```
michael-ciniawsky: This can be removed if `webpack-defaults > v0.4.4`

bebraw:
I would say merge, publish a 1.0 beta.

d3viant0ne:
@bebraw - A thought while I'm working on the exports-loader upgrade and it's tests, we had discussed testing everything through webpack which is something i'm doing in exports-loader.

We should really make it a point to either create an integration suite or just test everything through webpack to include file-loader.

bebraw:
@d3viant0ne Yeah, now that I think of it, testing directly through webpack is the sanest option. It adds some overhead but at least that avoids a category of bugs and integration tests are needed anyway so it's hard to avoid **not** running through webpack.

d3viant0ne:
Do we want to add webpack as a `peerDep` or just a note in the contribution guidelines that webpack is required for local development?

bebraw:
@d3viant0ne Since you are going to need it for development anyway, then a dev dep would make sense. Setting a peer dep as well would be good, though, as that would allow us to communicate version support to users (much harder to run incompatible versions).

d3viant0ne:
devDep limits our ability to test against multiple versions in travis

bebraw:
Good point. What if we push the problem to an init script a developer runs when starting to develop? That would do `npm i` and pull a fresh version of webpack to test against.

d3viant0ne:
Same issue applies to local development imo. Any automated install of Webpack will be limiting in the same manner.

I think particular instance, i think clear documentation should be enough. For anyone familiar with the ecosystem, needing Webpack for loader / plugin development is obvious. For anyone that isn't, it's a simple lesson they will only have to learn once which can be at least partially mitigated in a `DEVELOPMENT.md / CONTRIBUTING.md`

bebraw:
> I think particular instance, i think clear documentation should be enough. For anyone familiar with the ecosystem, needing Webpack for loader / plugin development is obvious. For anyone that isn't, it's a simple lesson they will only have to learn once which can be at least partially mitigated in a DEVELOPMENT.md / CONTRIBUTING.md

Ok, cool. 👍

bebraw:
@d3viant0ne: Do you want to move with this later? I don't mind updating the PR.

Issue #46, opened 2015-11-13
How to deal with relative public path


adcentury:
I set a relative public path in outputs

```
output: {
  path: path.join(__dirname, 'www/static'),
  filename: '[hash].bundle.js',
  publicPath: './'
}
```

Then the file structure is like

```
www
  |-- static
          | -- bundle.js
          | -- some-image.png
          | -- some-other-image.png
          | -- style.css
  |-- index.html
```

When i require image in css, i works fine.

But when i require image in some js, the url is wrong.

```
const url = require('path/to/some-image.png');
// the url will be 'some-image.png'
// but this url will be used in index.html (through react)
```

Any one can help me with it? Thanks.

silvenon:
Why don't you set it to `/` instead?

adcentury:
Because I want to use these files in phone gap app, using '/' will fail in it.

export-mike:
yeah same issue here @adcentury any progress?

dancoates:
It isn't a very good solution but you can work around this issue by placing the scripts and css in the www folder with the index.html and setting `publicPath : ""`. That way it doesn't matter if the files are accessed relative to the html or the css.

adcentury:
@export-mike The only way I found is to put all files (include html, js, css, images) into the www folder as @dancoates said. By the way, `HtmlWebpackPlugin` may help with generating html files.

mummybot:
Don't know if this helps, but I have the following:

```javascript
output: {
...
    publicPath: 'http://localhost:8080/path/to/build/'
},
devServer: {
   contentbase: 'path/to/build/'
```

mummybot:
Are you using these in a separate app? If so, have you tried setting up your webpack dev server to proxy the other app, that way you can use '/' as your publicPath base but still forward requests to other app e.g.

```javascript
var buildPath = '/wp-content/themes/mytheme/build/';

var config = {
  entry: [
    'webpack-dev-server/client?http://localhost:8080/',
    'webpack/hot/only-dev-server',
    path.join(__dirname, 'src/index.js')
  ],
  output: {
    path: path.resolve(__dirname, 'wp-content/themes/tpbc/build'),
    filename: 'index.js',
    publicPath: buildPath
  },
  devtool: 'eval-source-map',
  resolve: { alias: {} },
  devServer: {
    contentBase: buildPath,
    proxy: {
      '*': {
        target: externalServerURL,
        bypass: function(req) {
[...]

PR #135, opened 2017-03-09
feat: Emit files with relative urls

### LINKED ISSUE #62: Allow option for relative URLs
I have a scenario where I need the file-loader to return the url relative not absolute to the output path. Please allow an option to do this. Currently, I'm filtering out the pre-prended '/' with sed.

adriancmiranda:
A feat

The old tests should work.

The old docs should work too.

**Summary**

Emit files with relative urls
Fixes #62 and #46 

no


adriancmiranda:
@d3viant0ne I'm sorry for the sequence of 'commit'. I'll do it all soon. Thank´s!

adriancmiranda:
Done!

huan086:
Doesn't seem to handle my case

```js
module.exports = {
    entry: { "js/server": [ "./folder1/script1.jsx", "./folder2/script1.jsx" ] }
...
```

`./folder1/script1.jsx` has `require("../content/images/picture.png")` (relative to folder1)

The generated path is

`module.exports = __webpack_require__.p + "../content/images/content/images/picture.png";`

when I expected it to be

`module.exports = __webpack_require__.p + "../content/images/picture.png";` (relative to js)

adriancmiranda:
You can't use the [path] no the name

huan086:
@adriancmiranda nice catch!

There's another problem, which I don't know if it should be fixed by url-loader or file-loader or ExtractTextPlugin or webpack itself

My output folder is
```js
    output: {
        path: path.resolve(__dirname, ".")
```

In `./folder1/folder2/script1.jsx`, i have an `import "../../content/bootstrap.css";`

In `bootstrap.css`, it has
```css
@font-face {
  src: url('../fonts/glyphicons-halflings-regular.eot');
```
Folder structure
```
root
|-content
| |-bootstrap.css
| `-fonts
|   `-glyphicons-halflings-regular.eot
`-folder1
  `-folder2
    `-script1.jsx
```

It resulted in server.css
```css
src:url(../fonts/glyphicons-halflings-regular.eot)
```
which is invalid and `./fonts/glyphicons-halflings-regular.eot` being emitted

Expected output
Folder structure
```
root
|-content
| |-bootstrap.css
| `-fonts
|   `-glyphicons-halflings-regular.eot
|-folder1
| `-folder2
|   `-script1.jsx
`-js
  |-server.js
  `-server.css
```

I think for the issue 46 involving CSS, it can never be fixed as long as webpack don't handle CSS separately
https://github.com/webpack-contrib/file-loader/issues/46#issuecomment-196549110

adriancmiranda:
I don't know if I understand the problem(Mainly without view the webpack.config), you mean something like this?
https://bitbucket.org/adriancmiranda/pr-file-loader/src

huan086:
Wow! Close enough. In https://bitbucket.org/adriancmiranda/pr-file-loader/src/266ff6b5bd76b4f434cf21a61904f1198e36e1e5/content/bootstrap.css?at=master&fileviewer=file-view-default

change to `url("../fonts/bold-cyrillic-ext.woff2")` (double dots)

I'm thinking of using the following to fix
1. Change all URL in css to absolute path relative to root. e.g. `/content/fonts/bold-cyrillic-ext.woff2`
2. When it reaches `url-loader`, treat absolute path as relative to root, and modify the path to become relative to the javascript file containing the module.export. e.g. `<output directory>/server.js` would have `module.export = "@font-face { ... url(./content/fonts/bold-cyrillic-ext.woff2) ... }`
3. `file-loader` would emit files under `<output directory>/content/fonts/bold-cyrillic-ext.woff2`
4. ExtractTextPlugin, when configured to output the css to a directory other than `<output directory>/server.css`, would modify the `url` accordingly.

I'm new to webpack, no idea whether the above would work or not...

adriancmiranda:
Ok, let me see:

If you put `../fonts` (with double dots), the require method can't resolve this path because doesn't exists fonts directory on the root directory, this will generate an error, anyway, if I put the `fonts` directory in the `root`, generates an `outputPath` correct, but with a unexpected `url` in this specific case (pointing to a directory in the root).

*expected:*
`fonts`

*obtained:*
`../fonts` (without any change because the issuer context is pointing to up in relation to the font)

I'll fix this also validating the webpack public path parameter, not only outputPath (As I already do). Something like `path.relative(context, outputPath)` should be work, I think I can see about this later.

About the other things i.e. PostCSS and absolute paths

It solve but isn't the proposed solution.

adriancmiranda:
It works, but I have to find a way to write the tests yet.
https://bitbucket.org/adriancmiranda/pr-file-loader/src/27e9759b53f82c9e543b6749062f4acf84c2bbb5/loaders/file-loader.js?at=master&fileviewer=file-view-default#file-loader.js-59

d3viant0ne:
@adriancmiranda - Do you have anything else needing to be updated here pre you last comment or is this good to go?

adriancmiranda:
@d3viant0ne I prefer keep it simple for now, so, yes!

d3viant0ne:
@bebraw @michael-ciniawsky - Anything else before we we land this feature?

bebraw:
@d3viant0ne Nope. It's good to go.

adriancmiranda:
@huan086 I updated the sample folder, may you test for me again, please?
https://bitbucket.org/adriancmiranda/pr-file-loader/src
That should be working for your case too.

Jorybraun:
I am having the same issue, 
about the output paths in the parent folder.

adriancmiranda:
@Jorybraun This was already been solved, follow it [here](#150), please 😉

PR #139, opened 2017-04-01
fix: outputPath function overriden by useRelativePath

economysizegeek:

bugfix -  The new useRelativePath option code - did not handle the case where you used a function for the outputPath generation. 
Yes - I added in a test to make sure that if you use a function for outputPath it is called. If you use the useRelativePath it is ignored.
You don't currently document using outputPath as a function so I wasn't sure I should.
**Summary**
New code broke old behavior. If you use a function to set the outputPath it was expected that the function would be the decider on where the output of the file is.  Recent code took that away:(

Motivation: Recent code broke old behavior. I updated to the current version and my app stopped generating correctly.

No - it restores functionality.

**Other information**
There weren't any tests for outputPath as a function. I've added one so that in the future the feature will be protected.

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,7 +55,6 @@ module.exports = function(content) {
 			? config.outputPath(url)
 			: config.outputPath + url
 		);
-	}
 		url = outputPath;
 	} else {
 		outputPath = url;
--- a/index.js
+++ b/index.js
@@ -36,14 +36,6 @@ module.exports = function(content) {
 	});
 
 	var outputPath = "";
-	if (config.outputPath) {
-		// support functions as outputPath to generate them dynamically
-		outputPath = (
-			typeof config.outputPath === "function"
-			? config.outputPath(url)
-			: config.outputPath
-		);
-	}
 
 	var filePath = this.resourcePath;
 	if (config.useRelativePath) {
@@ -56,8 +48,14 @@ module.exports = function(content) {
 			outputPath = relativePath + url;
 		}
 		url = relativePath + url;
-	} else if (outputPath) {
-		outputPath = outputPath + url;
+	} else if (config.outputPath) {
+		// support functions as outputPath to generate them dynamically
+		outputPath = (
+			typeof config.outputPath === "function"
+			? config.outputPath(url)
+			: config.outputPath + url
+		);
+	}
 		url = outputPath;
 	} else {
 		outputPath = url;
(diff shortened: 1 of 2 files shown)


adriancmiranda:
@economysizegeek Ow! I see and I'm sorry for this, but this fix also broken the `outputPath` for `useRelativePath` at this point [#L46](https://github.com/webpack-contrib/file-loader/blob/master/index.js#L46)
Can you verify?

economysizegeek:
I left that block in tact and didn't modify it from what is happening with relative path. (I'm not sure it made sense to handle an outputpath function with the userelativepath - so i left them mutually exclusive).  Could you give a sample config and explain what happens that is wrong?

adriancmiranda:
Apparently not, but, if you use `path.join` foo with `../foo` you got  `foo` not `../foo`

economysizegeek:
Just so I could write a test - can you tell me the config.name you have set so I can see all three arugments sent to the join   so it resolves to path.posix.join("foo", "../foo", ?) == "../foo"

adriancmiranda:
@economysizegeek Sorry if I'm slow in answering. I'm working in another approach to `useRelativePath`. There are still outstanding issues at [#135](https://github.com/webpack-contrib/file-loader/pull/135). Honestly, I still don't know how to write the tests more consistently without using a "useless" property just to mock up the `issuerContext`(Maybe you can help me), but if you want to see the idea, follow the [link](https://github.com/adriancmiranda/file-loader/blob/master/index.js).

Jorybraun:
i'm still having this issue, when I use relative path. It overides the publicpath and appends '../' to the front of the relative folder.

`      {
          test: /\.(png|jpg|gif|svg|eot|ttf|woff|woff2)$/,
          loader: "file-loader?name=[name].[ext]",
          query: {
            useRelativePath: true
          }
      },`

Jorybraun:
`
      ../fonts/glyphicons-halflings-regular.eot  20.1 kB          [emitted]
                    ../img/exclamation-mark.png  2.13 kB          [emitted]
      ../fonts/glyphicons-halflings-regular.svg   109 kB          [emitted]
      ../fonts/glyphicons-halflings-regular.ttf  45.4 kB          [emitted]
     ../fonts/glyphicons-halflings-regular.woff  23.4 kB          [emitted]
    ../fonts/glyphicons-halflings-regular.woff2    18 kB          [emitted]
`

PR #145, opened 2017-04-05
fix: allow to override publicPath with an empty string

dms1lva:

Bugfix.
Yes.

**Summary**
When checking if the publicPath option was set, it would only do a falsy check which would return false with an empty string. Modified it to also set the public path if an empty string is given.

No.
**Other information**

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ module.exports = function(content) {
 	}
 
 	var publicPath = "__webpack_public_path__ + " + JSON.stringify(url);
-	if (config.publicPath || config.publicPath === "") {
+	if (config.publicPath !== false) {
 		// support functions as publicPath to generate them dynamically
 		publicPath = JSON.stringify(
 			typeof config.publicPath === "function"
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -61,7 +61,7 @@ module.exports = function(content) {
 	}
 
 	var publicPath = "__webpack_public_path__ + " + JSON.stringify(url);
-	if (config.publicPath) {
+	if (config.publicPath || config.publicPath === "") {
```
michael-ciniawsky: @dms1lva Could you try with `if (config.publicPath !== false)` like it's done in #122, I would favour this PR since it contains proper testing, but the fix in #122 is cleaner :)

dms1lva:
Made the change. My pleasure! :)

rtsao:
The webpack config validator does not allow `publicPath: false`. `undefined` is allowed, but not `false`. This change thus makes it impossible to cause `__webpack_public_path__` to be used.

bebraw:
@rtsao Can you open an issue so we don't lose the info and get it fixed? Thanks.

Issue #149, opened 2017-04-09
output images are in parent folder when useRelativePath option is true


nopantsmonkey:
My webpack file
```
import CleanWebpackPlugin from 'clean-webpack-plugin';
import StyleLintPlugin from 'stylelint-webpack-plugin';
import ExtractTextPlugin from 'extract-text-webpack-plugin';
import combineLoaders from 'webpack-combine-loaders';
import {modulesRoot, root, staticRoot} from './helpers';

const projectRootPath = root('.');
const modulesPath = modulesRoot();
const buildPath = staticRoot();

export default (env) => {
	let proMode = env.toLowerCase() !== 'production';

	return {
		devtool: proMode ? 'source-map' : 'eval',

		entry: {
			'css/web/home/styles': 'css/web/home/styles.pcss'
		},

		output: {
			path: buildPath,
			publicPath: buildPath,
			filename: proMode ? '[name].[chunkhash].css' : '[name].css',
			sourceMapFilename: proMode ? '[name].[chunkhash].map' : '[name].map'
		},

		resolve: {
			modules: [projectRootPath, modulesPath, root('images')],
			extensions: ['.pcss']
		},
		module: {
			rules: [
				{
					test: /\.(jpe?g|png|webp|svg|gif|cur)$/i,
					loader: 'file-loader',
					options: {
						name: '[path][name].[hash].[ext]',
						publicPath: '../../'
					}
				},
				{
					test: /\.pcss$/,
					loader: ExtractTextPlugin.extract({
						use: combineLoaders([
							{
								loader: 'css-loader',
								options: {
									importLoaders: 1,
									sourceMap: proMode,
									url: true
								}
							},
							{
								loader: 'postcss-loader',
								options: {
									sourceMap: proMode ? undefined : 'inline'
								}
							}
						])
					})
				}
			]
		},
		plugins: [
			...(env === 'watch' ? [] : [new CleanWebpackPlugin(['build'], {root: projectRootPath, verbose: true})]),
			new StyleLintPlugin({files: 'css/**/*.pcss'}),
			new ExtractTextPlugin({filename: (proMode ? '[name].[chunkhash].css' : '[name].css'), allChunks: true})
		],
		stats: {
			assets: true,
			children: false,
			chunks: true,
			chunkModules: true,
			chunkOrigins: true,
			colors: true,
			errors: true,
			errorDetails: false,
			hash: false,
			modules: false,
			reasons: false,
			source: true,
			timings: true,
[...]

PR #150, opened 2017-04-13
feat(index): add `cssOutputPath` option (`option.cssOutputPath`)

adriancmiranda:

A bugfix and a feature (`cssOutputPath`)

no

The readme file has been updated

**Summary**

#149 and #135 [public path validation](https://github.com/webpack-contrib/file-loader/pull/135#issuecomment-290170616)

no

**Other information**
Entry files and CSS output path doesn't pass through the `file-loader` so we haven't access to the file's context to compare with your issuer context correctly then we need to kidnap javascript entries and manually pass the CSS file output path using a new property (i.e. `cssOutputPath`). This should prevent unexpected output paths.


.gitignore
```diff
@@ -5,6 +5,8 @@
 
 # dotfiles
 .*
+!.editorconfig
 !.gitignore
+!.travis.yml
```
bebraw: Why to change these two?

index.js
```diff
@@ -1,80 +1,106 @@
 /*
-	MIT License http://www.opensource.org/licenses/mit-license.php
-	Author Tobias Koppers @sokra
+  MIT License http://www.opensource.org/licenses/mit-license.php
```
bebraw: The formatting change should probably be a separate PR.

.gitignore
michael-ciniawsky: Yep these will be take care of by #130

index.js
michael-ciniawsky:
```
npm i -D jsdoc-to-markdown
```

**package.json**
```json
{
  "scripts": {
     "docs": "jsdoc2md index.js > LOADER.md"
   }
}
```

**index.js**
```js
var path = require('path')
...
/**
 * File Loader
 *
 * > Loads and emits files
 *
 * @author Tobias Koppers (@sokra)
 * @version 0.12.0
 * @license MIT
 *
 * @module file-loader
 *
 * @requires path
 * @requires loaderUtils
 *
 * @method loader
 *
 * @param {String} file File
 *
 * @return {String} module Module (File)
 */
module.exports = function (file) {...}
```

index.js
```diff
@@ -1,80 +1,106 @@
...
-	var configKey = query.config || "fileLoader";
-	var options = this.options[configKey] || {};
-	var config = {
-		publicPath: false,
-		useRelativePath: false,
-		name: "[hash].[ext]"
-	};
+  var query = loaderUtils.getOptions(this) || {};
+  var configKey = query.config || "fileLoader";
+  var options = this.options[configKey] || {};
```
michael-ciniawsky: General Question: Is this still needed 😛  ? What does `this.options` contain ? 🙃

index.js
```diff
@@ -1,80 +1,106 @@
...
+  var config = {
```
michael-ciniawsky: `config` => `defaults` ?

index.js
```diff
@@ -1,80 +1,106 @@
...
+    useRelativePath: false,
+    name: "[hash].[ext]"
+  };
 
-	// options takes precedence over config
-	Object.keys(options).forEach(function(attr) {
-		config[attr] = options[attr];
-	});
+  // options takes precedence over config
+  Object.keys(options).forEach(function(attr) {
```
michael-ciniawsky: `Object.assign`?

index.js
```diff
@@ -1,80 +1,106 @@
...
+    config[attr] = options[attr];
+  });
 
-	// query takes precedence over config and options
-	Object.keys(query).forEach(function(attr) {
-		config[attr] = query[attr];
-	});
+  // query takes precedence over config and options
+  Object.keys(query).forEach(function(attr) {
```
michael-ciniawsky: `Object.assign`? Why is that done twice ?

index.js
```diff
@@ -1,80 +1,106 @@
...
-	if (config.publicPath) {
-		// support functions as publicPath to generate them dynamically
-		publicPath = JSON.stringify(
-			typeof config.publicPath === "function"
-			? config.publicPath(url)
-			: config.publicPath + url
-		);
-	}
+    // We have access only to entry point relationships. So we work with this relations.
+    var issuerContext = (this._module && this._module.issuer && this._module.issuer.context) || context;
```
michael-ciniawsky:
```js
const issuer = {...}
const relative = { url: '...' , path: '...' }
const output = {...}
```

index.js
```diff
@@ -1,80 +1,106 @@
...
+    // So we don't have access to the files context to compare with your assets context
+    // then we need to kidnap from `extract-text-webpack-plugin` if you using him and the same way
+    // force the `relativePath` to extract files on the webpack output path config folder.
+    // otherwise make the same using the same output path from text extracted
+    // with `extract-text-webpack-plugin`.
+    var output = this.options.output || {};
+    if (output.filename && output.filename !== "extract-text-webpack-plugin-output-filename") {
+      relativePath = outputDirname;
+    } else if (toString.call(config.textOutputPath) === "[object String]") {
+      var outputPackageDirname = output.path.replace(this.options.context + path.sep, "");
```
michael-ciniawsky: See the comment directly above, please group `issuer, relative, output` via an `{Object}`

michael-ciniawsky:
Is this in general 'only' related to `extract-text-webpack-plugin`, or could this be used by other plugins etc aswell? `assetsOutputPath || assetsPath`

index.js
adriancmiranda: @bebraw I agree. I'll change this.
@michael-ciniawsky I don't know if I can change the header. I see a pattern in other plugins

index.js
michael-ciniawsky: kk leave it :) I will test it when time and it's separate PR

index.js
michael-ciniawsky: kk

index.js
michael-ciniawsky:
```js
      const issuer = {}
      const output = this.options.output || {}
      const relative = {}

      issuer.ctx = (this._module && this._module.issuer && this._module.issuer.context) || context;

      relative.url = issuer.ctx && path.relative(issuer.ctx, this.resourcePath);
      relative.path = relative.url && path.dirname(relativeUrl);

      output.dirname = relative.path
        .replace(/\.\.(\/|\\)/g, "")
        .split(path.sep)
        .join("/");

     if (output.dirname.indexOf(outputPath) !== 0) output.dirname = outputPath;

     outputPath = path.join(output.dirname, url)
          .split(path.sep)
          .join("/");
 
    
    if (output.filename && output.filename !== "extract-text-webpack-plugin-output-filename") {
       relative.path = output.dirname;
    } else if (toString.call(config.textOutputPath) === "[object String]") {
      output.ctx= output.path.replace(this.options.context + path.sep, "");
      output.asset = path.join(context, output.ctx, output.dirname);

      issuer.output = path.join(context, output.ctx, config.textOutputPath);

      relative.path = path.relative(issuer.output, output.asset);
    }
      url = path
       .join(relative.path, url)
       .split(path.sep)
       .join("/");
    } else if (config.outputPath) {
      url = outputPath;
    } else {
      outputPath = url;
    }
```

It about grouping the various values/variables together 😛

michael-ciniawsky:
>  extract-text-webpack-plugin-output-filename is sent for all requested files (except js entries) by the specific plugin, but the output.filename is from webpack.config then I don't think that's a problem to use with another plugin unless it also sends something like another-plugin-output-filename for all requested files.

kk, yeah this might be out-of-scope of this PR, I'm sry to bikeshed about this, but we need to make sure to **not** let any specific hooks relying on other plugins/loaders into loaders/plugins in general when possible and try go for the generic solution first. So if we can design it as a generic hook for plugins in need of the same functionality as ETWP, we should do it. 

But this doesn't mean it's a blocker for this PR :), tbh I have no clue if any other plugin would need this aswell. I just need to clearify this 😛 

> The textOutputPath is relative to the extracted file (e.g css), is from the issuer, not for your assets. I'm not sure if the extract-text(for example) is used only for CSS files so I prefer to keep the generic name or maybe cssOutputPath. I don't know. What do you think?

If we go with this, we need to make sure folks using the `file-loader`+ ETWP combo, directly see this option is for you 😛 

`textOutput[Path]` (Current)
`extractOutput[Path]`
`extractTextOutput[Path]`

index.js
michael-ciniawsky: ? 🙃

adriancmiranda:
@michael-ciniawsky No, you're totally right and I'm happy to answer.
Actually, my idea is quite simple and maybe that's exactly why it's confusing without seeing it in action.
At [this](https://github.com/adriancmiranda/file-loader/blob/master/index.js#L75) line, I expected only JavaScript files and I filled it with [this](https://github.com/adriancmiranda/file-loader/blob/master/index.js#L56) info because it has only the part of the relative path that matters i.e without any ../ because it's generally requested from the root. So actually [this](https://github.com/adriancmiranda/file-loader/blob/master/index.js#L74) line is only preventing for that ETWP overrides any path with `extract-text-webpack-plugin-output-filename` (not is a dependency) and bring to us the javascript path too, to kill two birds with one stone. Already on [this](https://github.com/adriancmiranda/file-loader/blob/master/index.js#L80) line I particularly is expecting only CSS files but nothing prevent it from being another kind of file (the only reason I called from `textOutput` and not `cssOutput`), again, you're right, this name generates a mess and my quotations for the plugin enforce this. But how can we call it?

index.js
d3viant0ne: Stick a todo in there to switch this to `Object.assign` in the defaults pull request which removes support for older node version.

index.js
```diff
@@ -66,12 +66,10 @@ module.exports = function(content) {
...
 		// So we don't have access to the files context to compare with your assets context
-		// then we need to kidnap from `extract-text-webpack-plugin` if you using him and the same way
-		// force the `relativePath` to extract files on the webpack output path config folder.
-		// otherwise make the same using the same output path from text extracted
-		// with `extract-text-webpack-plugin`.
+		// then we need to create and the same way force the `relativePath` to bundled files
+		// on the webpack output path config folder and manually the same with CSS file.
 		var output = this.options.output || {};
-		if (output.filename && output.filename !== "extract-text-webpack-plugin-output-filename") {
+		if (output.filename && path.extname(output.filename) === ".js") {
```
michael-ciniawsky: 👍

adriancmiranda: It's wrong yet. I remembered, it has 'jsx' too, maybe another kind of file.

adriancmiranda:
@michael-ciniawsky That's it!
I updated the samples too

https://bitbucket.org/adriancmiranda/pr-file-loader/src

https://bitbucket.org/adriancmiranda/pr-file-loader/src/489f5e70002496b4098a283b1a794643ee0869e7/?at=file-loader-relative-path-issue

michael-ciniawsky:
- `cssPath || stylePath` is better imho, but that's majority decision*
- Docs (more) 😉 
- Tests

Also see https://github.com/webpack-contrib/file-loader/pull/150#discussion_r111624781

* we maybe even consider to put this under one option prop in the future `paths: { public: '', css: '', ...}`, but not yet...

index.js
adriancmiranda: It makes sense for me, but I also think it's necessary another PR

michael-ciniawsky: Why ? 🙃

adriancmiranda: To separate the things and to avoid break changes also my last commit generated the #139.
I think in a style PR it's more appropriate

michael-ciniawsky: Nope not really 😛 , but it's ok to leave it for now

Cap32:
I have the same problem, and this pr resolve my problem. Hope to merge it. 👍🏿

index.js
adriancmiranda: Sry I didn't understand before, I thought it was part of the review ^^

index.js
adriancmiranda: @michael-ciniawsky So, let's do it!

index.js
adriancmiranda: @d3viant0ne Can I make this change in this PR?

index.js
adriancmiranda: So, let's do it!

adriancmiranda:
@michael-ciniawsky 

> cssPath || stylePath is better imho, but that's majority decision*
I don't know. Particularly I prefer to keep explicit the `output` term.

> * we maybe even consider to put this under one option prop in the future paths: { public: '', css: '', ...}, but not yet...

I like it and I can make it too if you wish.
Check also for me if you'll want this in this PR: https://github.com/webpack-contrib/file-loader/pull/150#discussion_r111669017

index.js
michael-ciniawsky: In the gist a newline `\n` after the follwing lines and I'm 👍 on Code Style 😛 L50, 65, 74, 84

michael-ciniawsky:
I merged a small bugfix please rebase :), 

> I don't know. Particularly I prefer to keep explicit the output term.

`cssOutputPath` is verbose, but like a metioned, just my personal opinion and majority decides, so just leave it for now 😉 

> we maybe even consider to put this under one option prop in the future paths: { public: '', css: '', ...}, but not yet...

This is a sadly a breaking change, I need to stop thinking out loud in PR's, it's confusing for others 😛

adriancmiranda:
Sorry! I did it wrong 😨

michael-ciniawsky:
`git reset --hard HEAD~1 && git checkout master && git fetch upstream master && git merge master && git checkout file-loader && git rebase master`

adriancmiranda:
@michael-ciniawsky I made the terrible mistake of making the changes directly in the `master branch` of my fork as the last time
so I used `git pull --rebase origin master`

michael-ciniawsky:
@adriancmiranda I'm sry 😛  I thought merging the smaller PR would be the simpler solution instead of the otherway around. Maybe just copy and paste your changes and open a new clean PR ?

adriancmiranda:
@michael-ciniawsky [What about it?](https://github.com/webpack-contrib/file-loader/pull/150#discussion_r111669017)

michael-ciniawsky:
Onyl the style related one's you posted in the gist, othwerwise should be fine then imho, but we need to wait for th revei of others aswell :)

adriancmiranda:
You ar right, well, thank's for your review. 😉

index.js
adriancmiranda: up-to-date

michael-ciniawsky: @adriancmiranda Update this PR with the Gist, we will squash it later anyways

README.md
```diff
@@ -61,6 +61,8 @@ use: "file-loader?name=[name].[ext]&publicPath=assets/foo/&outputPath=app/images
 }
 ```
 
+`cssOutputPath` also should be used (together with `useRelativePath` property) to define the output path context.
```
michael-ciniawsky: Please add e.g an example for this awell :). The current explanation is too jaded to reason about it (especially for beginners)

adriancmiranda: hahah, ok, I'll make a sample folder and write some tests on it. I don't know if I finish today but I'll try. And, yes, I need to put the webpack and ETWP, also make some fixes on travis, maybe some `commits` to fix it or put engines on the package to fix the integration test. Is there any problem?

michael-ciniawsky:
```
|– src
|    |– css
|         | ...
|– dist 
|     |– bundle.js
|     |– bundle.css
|– webpack.config.js
|– package.json
```

**webpack.config.js**
```
...

{
  loader: 'file-loader'
  options: {
     useRelativePath: true,
     cssOutputPath: '...'
   }
}
...

{
    use: ExtractTextPlugin.extract({
      fallback: 'style-loader'
      use: [ 'css-loader' ]
    })
}
...

new ExtractTextPlugin()
```

**path/to/file.(svg|png|gif|jpg)**
```
// Example Asset
```

**path/to/file.css**
```
// Example CSS
```

**bundle.css**
```
// CSS Bundle (cssOutputPath + ETWP)
```

I came back here later in terms of tests 😛

adriancmiranda: Ok, I didn't saw your message before, so I made a little different :)

michael-ciniawsky: The example is 👍 I know it's annoying but we need to get this into the docs aswell 😛

michael-ciniawsky:
@bebraw @d3viant0ne Please general approval/rejection on this feature, before we continue for nothing in the end :)

README.md
adriancmiranda: No, that's fine; It's fair, but I ain't finished yet. ^^
I'll come back later and then I try to write something better.

michael-ciniawsky: Merging them would be a breaking change, please leave it for the moment I need to think about how to handle it at best first 😛 it maybe better a separate PR in conjunction with `webpack-defaults` upgrade and another semver majors e.g the general paths unification like I mentioned before

adriancmiranda: **README**
```
The CSS files don't pass through the `file-loader` when they are extracted from `webpack`.
So, at this moment, we haven't access to your contexts to be compared with your own resources context to after that we generate a relative path to them.
So we should pass it manually with `cssOutputPath`.
This should prevent unexpected output paths to your own CSS assets.
See the examples folder to see it working.
```
Is it better?

.github/assets/file_loader_icon.svg
```diff
@@ -1,36 +1,36 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="200px" height="200px" viewBox="0 0 200 200" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
bebraw: Why to change this icon?

adriancmiranda: @bebraw Sorry, are you using it?
I changed just to show in an example. I looked for references to him before, I didn't find so
I didn't think it could be a problem.

d3viant0ne:
@michael-ciniawsky - I don't have an issue with this conceptually nor with the quality of the pull request. There is a lot of noise here so I'll have to wait to review it until later this evening.

michael-ciniawsky:
@d3viant0ne We definitely need to check this again, since it is a conjunction between ETWP && `file-loader` to some point which is always undesireable, but seems to be no better way to solve the particular issue atm. I hadn't time to play with this maybe later this week 😛

d3viant0ne:
Yeah, I just saw that. That's probably going to be a deal breaker if we can't find a way to solve this problem without depending on another loader.

When in doubt, wait.

adriancmiranda:
Guys, sorry, maybe I was not clear, my english isn't good
but, you can use with another loaders too e.g. [extract-loader](https://www.npmjs.com/package/extract-loader). In fact, we need only to know where the css will be saved.

thefabulousdev:
Any update on this?

d3viant0ne:
@adriancmiranda - - I have set this to blocked ( #165 ) & moved this feature into `1.1.0`, the next minor range after the pending `1.0.0` release.

adriancmiranda:
@d3viant0ne Thank you... and I'm sorry for the noise :wink:

stereokai:
@adriancmiranda @d3viant0ne

What about this solution? Will it ever be merged? Half of my images end up in the parent folder.

Thanks!

adriancmiranda:
Hey @stereokai I'm waiting for another review yet. 😉

stereokai:
@adriancmiranda Actually, here's an update: 

After 2 days of trial and error (Thanks Webpack docs!) I found out that there's a way to control file-loader's output and replicate the file structure of the source directory in the output directory. It's the Webpack setting `context`, which can be used per loader or per build, For file-loader:

      use: [{
        loader: 'file-loader',
        options: {
          context: <path>, // The directory to draw relative paths from
          name: '[path][name].[ext]'
        },
      },

Here's an example of how to use it. Let's say your desired result is to load an image via `www.website.com/assets/images/user.png`

Your project's file structure is:

    project/
    ├── src/
    │   └── assets/
    │       └── images/
    │           └── user.png
    └── build/

And the desired result is:

    project/
    ├── src/
    │   └── assets/
    │       └── images/
    │           └── user.png
    └── build/
        └── assets/
            └── images/
                └── user.png

The configuration for this is:

      use: [{
        loader: 'file-loader',
        options: {
          context: path.resolve(__dirname, 'src')
          name: '[path][name].[ext]'
        },
      },

And that is because you want to replicate the file structure under `src` folder inside the build `folder`.

If you wanted to remove the `assets` directory from the path, the value for `context` would be: `path.resolve(__dirname, 'src/assets')`, and file-loader will replicate the relative paths starting in 'src/assets' instead, and the result would be: 

    project/
    ├── src/
    │   └── assets/
    │       └── images/
    │           └── user.png
    └── build/
        └── images/
            └── user.png

adriancmiranda:
@stereokai How about the links?

stereokai:
@adriancmiranda what do you mean? I was getting build assets in parent folders of the project root. Are we not talking about the same problem?

adriancmiranda:
@stereokai What I'm proposing isn't exactly solving a problem, but provide an additional solution to, so that you can also have relative paths in the compiled version of your code. Your example works perfectly, however, it only has one file-per-configuration context. So let's say you have this structure:

```javascript
src/
│   └── scripts/index.js -> require('../assets/images/placehold.jpg') 
│   └── assets/
│       └── images/
```

you get something like this, right?

`/src/assets/images/placehold.jpg`

My proposal is that you have a result like this:

`../assets/images/placehold.jpg`

considering of course that you'll replicate `src` dir. But let's say you want to change the build path from the `js` file to the root, something like this.

```javascript
build/
│   └── index.js -> require('../assets/images/placehold.jpg') 
│   └── assets/
│       └── images/
```

In this case you should have this path

`./assets/images/placehold.jpg`

instead

`/assets/images/placehold.jpg`

stereokai:
@adriancmiranda Does this line: `scripts/index.js -> require('../assets/images/placehold.jpg')` mean `scripts/index.js` is importing `placehold.jpg`?

adriancmiranda:
@stereokai Yes, it is, sorry about it

stereokai:
@adriancmiranda I'm sorry but I don't understand what it is you are trying to achieve

adriancmiranda:
@stereokai If you are curious about this, I recommend to you test [this](https://github.com/webpack-contrib/file-loader/pull/150#issuecomment-336943061) version. I would like to hear your feedback 😉

Leward:
The solution worked on my project (Cordova project so I cannot use `/` as public path)

In my package.json I put: 

```
"file-loader": "git://github.com/woshi82/file-loader.git#master"
```

In my webpack config: 

```
modules: {
  rules: [
    // ...
    {
        test: /\.(png|jpe?g|gif|svg)(\?.*)?$/,
        loader: 'url-loader',
        options: {
          limit: 5000, 
          name: utils.assetsPath('img/[name].[hash:7].[ext]'),
          useRelativePath: process.env.NODE_ENV === "production",
          cssOutputPath: 'static/css'
        }
      }
  ]
}
```

adriancmiranda:
@Leward The `useRelativePath` should works on development mode too :-)

PR #157, opened 2017-04-30
feat(index): add context to `outputPath` && `publicPath`

lsycxyj:
I propose to add issuerContext to outputPath and publicPath to allow users change the resolved paths dynamically (eg. implementing relative path by myself).
BTW,  the outputPath should be able to be changed by the users after using "useRelativePath" because it will emit files to wrong paths (eg. files will be placed to parent folder of output path when it starts with "..").

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,8 +38,8 @@ module.exports = function(content) {
 	var outputPath = "";
 
 	var filePath = this.resourcePath;
+	var issuerContext = this._module && this._module.issuer && this._module.issuer.context || context;
 	if (config.useRelativePath) {
-		var issuerContext = this._module && this._module.issuer && this._module.issuer.context || context;
 		var relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join("/");
 		var relativePath = relativeUrl && path.dirname(relativeUrl) + "/";
 		if (~relativePath.indexOf("../")) {
@@ -48,11 +48,12 @@ module.exports = function(content) {
 			outputPath = relativePath + url;
 		}
 		url = relativePath + url;
-	} else if (config.outputPath) {
+	}
+	if (config.outputPath) {
 		// support functions as outputPath to generate them dynamically
 		outputPath = (
 			typeof config.outputPath === "function"
-			? config.outputPath(url)
+			? config.outputPath(url, issuerContext)
 			: config.outputPath + url
 		);
 		url = outputPath;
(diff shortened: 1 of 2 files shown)


lsycxyj:
Knock, knock. Nobody's here?

adriancmiranda:
@lsycxyj This will being resolved in #150 😉

lsycxyj:
@adriancmiranda Thank you. I didn't mean that I only wanna fix the bug, but also wanna let the users handle the paths by themselves.

michael-ciniawsky:
@lsycxyj Sry we are shaping up a few things behind the scences in terms of CSS handling and this likely affects `file/url-loader` aswell, so they get stalled a bit atm, please be patient 😛

lsycxyj:
@michael-ciniawsky I'm looking forward to your good news. : )

michael-ciniawsky:
@lsycxyj Can you please provide a small example for this, just to clarify ? :)

index.js
```diff
@@ -38,8 +38,8 @@ module.exports = function(content) {
 	var outputPath = "";
 
 	var filePath = this.resourcePath;
+	var issuerContext = this._module && this._module.issuer && this._module.issuer.context || context;
```
michael-ciniawsky: nitpick `issuerContext` => `issuer || context` (personally `context`)

bebraw: Yeah, better clean this up.

lsycxyj: @michael-ciniawsky Well... Actually this line of code is from the `master` branch. I just moved it outside the block. Do you mean I should choose `issuer` or `context` to pass to the `publicPath` function?

lsycxyj: @michael-ciniawsky My purpose of adding issuer's context to the `publicPath` function is to give the   user a way to the resolved path trickily (eg. I can give a new relative path according to the issuer's context). So I prefer the issuer's context personally.

But I didn't fix the "useRelativePath" bug in this patch.

I notice that all the test cases are mocked cases, and there's no (good) way to find out problems caused by the webpack APIs (eg. `this._module`).
I'm not familiar with the webpack APIs. Maybe I need some help or advice to test it in a better way.

BTW, `this._module` has been deprecated by [officual docs](https://doc.webpack-china.org/api/loaders/#this-_module). Maybe it should be changed in some other way?

d3viant0ne:
@lsycxyj - I have set this to blocked ( #165 ) & moved this feature into `1.1.0`, the next minor range after the pending `1.0.0` release.

evilebottnawi:
@lsycxyj friendly ping

lsycxyj:
@evilebottnawi Is there anything I need to do?

evilebottnawi:
Done in https://github.com/webpack-contrib/file-loader/pull/305

lsycxyj:
My idea was merged into the library but I'm not in the contributors list. :cry:

evilebottnawi:
@lsycxyj what do you mean?

PR #159, opened 2017-05-05
fix: init `publicPath` to undefined

### LINKED ISSUE #158: regression: publicPath can't be false
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

The webpack config validator does not allow publicPath: false. undefined is allowed, but not false. This change thus makes it impossible to cause __webpack_public_path__ to be used.

**If the current behavior is a bug, please provide the steps to reproduce.**

See: https://github.com/webpack-contrib/file-loader/pull/145

**What is the expected behavior?**

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please mention other relevant information such as your webpack version, Node.js version and Operating System.**

d3viant0ne:

bugfix

**Summary**
Initializing `publicPath` to false was preventing the new option from being used with `webpack-config-validator
`

See comment in #145

No

**Other information**

Fixes #158

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@ module.exports = function(content) {
 	var configKey = query.config || "fileLoader";
 	var options = this.options[configKey] || {};
 	var config = {
-		publicPath: false,
+		publicPath: undefined,
 		useRelativePath: false,
 		name: "[hash].[ext]"
 	};
@@ -61,7 +61,7 @@ module.exports = function(content) {
 	}
 
 	var publicPath = "__webpack_public_path__ + " + JSON.stringify(url);
-	if (config.publicPath !== false) {
+	if (config.publicPath !== undefined) {
 		// support functions as publicPath to generate them dynamically
 		publicPath = JSON.stringify(
 			typeof config.publicPath === "function"


d3viant0ne:
It already has two to cover the feature in question, just changing how we init publicPath because of the issue mentioned in the PR that broke it.

bebraw:
Ok, cool. JS. 👍

index.js
```diff
@@ -61,7 +61,7 @@ module.exports = function(content) {
 	}
 
 	var publicPath = "__webpack_public_path__ + " + JSON.stringify(url);
-	if (config.publicPath !== false) {
+	if (config.publicPath !== undefined) {
```
michael-ciniawsky:
```js
if (!config.publicPath) {}
```

Maybe even 'safer', but enterily a nitpick

jcperez-ch:
Any news regarding this bug?

michael-ciniawsky:
@d3viant0ne  @bebraw Is there anything left to do here/blocking this ? :)

bebraw:
@michael-ciniawsky Not that I know of.

michael-ciniawsky:
Please semver patch someone when time :)

PR #165, opened 2017-06-05
feat: Apply webpack-defaults v1.x

d3viant0ne:
**What kind of change does this PR introduce?**

**Refactoring**

**Did you add tests for your changes?**

**If relevant, did you update the README?**

**Summary**

 - Retains index.js file history
 - Assumes proposed change to npm@5
 - Assumes proposed appveyor proposal

Can't rebase in the changes from master given the `index.js` in the old defaults upgrade was deleted. 

Has the latest proposed changes to defaults ( appveyor & npm@5 )

**Does this PR introduce a breaking change?**

_Breaking Change_ - Enforces NodeJS > 4.3 via engines.

Must release as semver `MAJOR`

**Other information**

PR #166, opened 2017-06-06
feat: add context to `options.outputPath` && `options.publicPath`

coot:
The context includes three fields userRequest, rawRequest and issuer context.

Adds functionality to outputPath and publicPath options (when set to functions)

Yes.

No, not yet, waiting for comments.

**Summary**
I am working on an react app in which I need to have relative path for files.  The static assets are served from `./static` folder.  When the app is requesting images there are two cases:
* they are requested from javascript bundle they need to have publicPath set to "./static/"
* if they are requested from "style.css" which by itself is inside "./static" folder they need to be relative to "./" rather than "./static"

No.

**Other information**

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,12 +50,9 @@ module.exports = function(content) {
 		url = relativePath + url;
 	} else if (config.outputPath) {
 		// support functions as outputPath to generate them dynamically
-                var userRequest = this._module && this._module.userRequest
-                var rawRequest = this._module && this._module.rawRequest
-                var context = this._module && this._module.issuer && this._module.issuer.context || context;
 		outputPath = (
 			typeof config.outputPath === "function"
-			? config.outputPath(url, { userRequest, rawRequest, context })
+			? config.outputPath(url, this._module || {})
 			: config.outputPath + url
 		);
 		url = outputPath;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -50,9 +50,12 @@ module.exports = function(content) {
 		url = relativePath + url;
 	} else if (config.outputPath) {
 		// support functions as outputPath to generate them dynamically
+                var userRequest = this._module && this._module.userRequest
```
bebraw: It would be neater to do

```javascript
var module = this._module || {};
var userRequest = module.userRequest;

...
```

d3viant0ne:
@coot - I have set this to blocked ( #165 ) & moved this feature into `1.1.0`, the next minor range after the pending `1.0.0` release.

lsycxyj:
@coot Do you have any good idea to test this, especially things like `this._module.issuer` from Webpack API?

coot:
@michael-ciniawsky I added a simple test that checks that a second argument was passed to the `publicPath` and `outputPath` callbacks.
@lsycxyj  I don't know how to test the issuer.

evilebottnawi:
Close in favor https://github.com/webpack-contrib/file-loader/pull/166. I think better pass `this` instead each property (more universally).

PR #178, opened 2017-07-14
fix: remove `=` from default export (`SyntaxError`)

jungomi:
The refactoring to ES modules (#167) accidentally left an `=` in the default export, which is a syntax error.


bastimeyer:
Could you please bump the version and publish a new release with this bugfix? Thanks!

PR #183, opened 2017-07-27
fix(cjs): export `raw` value

evilebottnawi:
We use `"main": "dist/cjs.js"` in `package.json`, but `default`(https://github.com/webpack-contrib/file-loader/blob/master/src/cjs.js#L1) doesn't contain `raw` value so `file-loader` doesn't work correctly.

### DIFF
--- a/src/cjs.js
+++ b/src/cjs.js
@@ -1,1 +1,4 @@
-module.exports = require('./index').default;
+const loader = require('./index');
+
+module.exports = loader.default;
+module.exports.raw = loader.raw;
--- a/test/cjs.test.js
+++ b/test/cjs.test.js
@@ -1,0 +1,12 @@
+import fileLoader from '../src';
+import cjsFileLoader from '../src/cjs';
+
+describe('cjs', () => {
+  it('should export loader', () => {
+    expect(cjsFileLoader).toEqual(fileLoader);
+  });
+
+  it('should export `raw` flag', () => {
+    expect(cjsFileLoader.raw).toEqual(true);
+  });
+});


src/cjs.js
```diff
@@ -1 +1,4 @@
-module.exports = require('./index').default;
+const fileLoader = require('./index');
```
michael-ciniawsky: `fileLoader` => `loader` 😛

src/cjs.js
```diff
@@ -1 +1,4 @@
...
+
+module.exports = fileLoader.default;
+module.exports.raw = fileLoader.raw;
```
Kovensky: This is mutating `fileLoader.default`... 😕

Maybe it could be better to write the workaround in `index.js` itself, with a comment that it's a workaround for the commonjs export.

They'll both do the same thing, and I guess technically put the cjs change outside of the cjs wrapper, but I'd prefer that to being mutated from outside.

evilebottnawi: @Kovensky Perhaps, but on the other hand, I would not want to add workarounds to the loader code, because it can be misleading. In any case, I'm open to ideas, I'll be glad to see another solution

src/cjs.js
evilebottnawi: @michael-ciniawsky :+1: done

src/cjs.js
michael-ciniawsky: @Kovensky What are potential issue with mutating `loader.default` ? Handling the Module System related stuff in the wrapper would 👍  if possible 🙃

evilebottnawi:
@michael-ciniawsky just ensure tests are good, also npm have bug with `node` 4.3 :disappointed: https://github.com/npm/npm/issues/18395

brianhelba:
@evilebottnawi See #204 for a fix to tests.

michael-ciniawsky:
@brianhelba This needs to be updated in `webpack-defaults` instead as it is the boilerplate for all repos

@d3viant0ne @evilebottnawi If it is the only known failure blocking, can we merge the fix regardless of the npm CI incompat for now and fix it in a follow up PR? Setting `raw` is fairly important for the loader

brianhelba:
@michael-ciniawsky I agree, I'd like to merge this PR ASAP, even without the fix in #204. The CI tests are already failing on `master`, so this won't make anything worse (and this still passes tests on everything except Node.js 4.3). However, this does fix a very serious regression, which makes `file-loader` totally unusable for binary files (e.g. images) right now.

@d3viant0ne @evilebottnawi Is that ok?

d3viant0ne:
@brianhelba - `file-loader@1.1.2`

brianhelba:
Excellent! Thanks for the quick attention and release.

PR #189, opened 2017-08-11
fix(index): Add check for `this.options`

jquense:
`this.options` doesn't seem to be defined always. If I don't have the LoaderOptionsPlugin it crashes :/


src/index.js
```diff
@@ -10,7 +10,7 @@ export default function fileLoader(content) {
 
   const query = loaderUtils.getOptions(this) || {};
   const configKey = query.config || 'fileLoader';
-  const options = this.options[configKey] || {};
+  const options = this.options && this.options[configKey] || {};
```
evilebottnawi: Why? please provide more information, also need tests, thanks!

jquense: I'm not sure what I would write a test for, without setting up and running throw webpack. The loader assumes this.options is provided by webpack unconditionally its not. It only exists in v2+ if the user provides the LoaderOptionsPlugin, a plugin that exists only to help ease migration from vq. This is just simple mistake on the webpack api. Either this or add to the Readme saying the LoaderOptionsPlugin is required

evilebottnawi: @jquense i am use `file-loader` without `LoaderOptionsPlugin`, can your provide repo where it is not works?

jquense: Sure.

jquense:
Sorry about that, the problem is that if you _do_ provide a LoadersOptionsPlugin but leave options undefined. On the face of it it seems like a tiny edge case. but we generally use functions to build our configs so this is actually easy to do e.g.

```js
getLoaderOptions = (options) => new webpack.LoaderOptionsPlugins({
  minimize: true, 
  options
})
```

.gitignore
```diff
@@ -15,4 +16,4 @@ Thumbs.db
 .idea
 .vscode
 *.sublime-project
-*.sublime-workspace
\ No newline at end of file
+*.sublime-workspace
```
evilebottnawi: Please exclude this from commit, we use `webpack-defaults` for all repos, need do PR there https://github.com/webpack-contrib/webpack-defaults

jquense: Sorry I'm not sure what you want excluded here, the line was already there my editor just inserted a newline character. I'm not even using sublime

michael-ciniawsky:
@jquense In case you are using webpack >= v2.0.0 please setup options in module.rules instead, since `LoaderOptionsPlugin` main purpose was to support **incompatible** loaders with webpack >= v2.0.0` during webpack`v2.0.0-beta` and as migration path. It is not needed anymore for most loaders (e.g `file-loader`). `this.options` (Mutliple Configs) is going away hopefully soon (#184) since usage is deprecated

evilebottnawi:
@michael-ciniawsky can we close this PR?

PR #213, opened 2017-09-30
feat(index): cache `outputPath` to avoid remitting files

### LINKED ISSUE #55: How to skip files rebuild  everytime the app runs in dev mode?
Is there anyway I can skip the files being rebuild every time the build runs? I don't see why I would want that in files that haven't change. Am I missing something?
Thanks :+1:

michael-ciniawsky:
### Notable Changes
---

Simple Memory Cache to avoid remitting files. Needs triage and testing though

### Issues
---

- Fixes #55

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -62,7 +62,15 @@ export default function loader(content) {
     );
   }
 
+  // Memcache output paths (files)
+  const cache = new Set();
+
   if (options.emitFile === undefined || options.emitFile) {
+    // Don't emit files again if cached
+    if (cache.has(outputPath)) return null;
+
+    cache.add(outputPath);
+
     this.emitFile(outputPath, content);
   }
   // TODO revert to ES2015 Module export, when new CSS Pipeline is in place
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,19 @@
 {
   "name": "file-loader",
-  "version": "1.1.0",
+  "version": "1.1.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "dev": true,
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "abab": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -62,7 +62,14 @@ export default function loader(content) {
     );
   }
 
+  // Memcache output files
+  const cache = {};
```
evilebottnawi: Maybe best use `Set`

d3viant0ne: Agreed, most benchmarks show set to be 2 / 3 times faster than the old school object route.

d3viant0ne: I think we are at the point where we can start being a bit more picky about performance related choices

michael-ciniawsky: Done

michael-ciniawsky: Done

src/index.js
```diff
@@ -62,7 +62,15 @@ export default function loader(content) {
...
   }
 
+  // Memcache output paths (files)
+  const cache = new Set();
+
   if (options.emitFile === undefined || options.emitFile) {
+    // Don't emit files again if cached
+    if (cache.has(outputPath)) return null;
+
+    cache.add(outputPath);
```
evilebottnawi: What about if file was changed? Not emitted new?

michael-ciniawsky: Yep, with the current solution you need to restart, this is a trivial guard atm to avoid remitting 'static assets' constantly when changing e.g CSS files. Should we really hash them ? Is there a low overhead solution to do so ?

evilebottnawi: @michael-ciniawsky i think `hash` is best solution for this, or maybe we can use `memory-fs` which build-in `webpack` for this purpose? Need to ask @sokra about using `memory-fs` (or other cache mechanism) into `loader`

michael-ciniawsky: I take a look while writing test, if there is a hash available somewhere and try to use it as the identifier

d3viant0ne: Curious if this is going to be an net gain or a net loss performance wise considering we would have to introduce hashing into the mix once you start throwing entire applications at it.

michael-ciniawsky: Yep, if their is a hash on the `module` `{Object}` anyways (need to check it) or another prop I can use as identifier, fine. But pulling e.g `crypto` in and generate a hash just for this trivial cache, it could likely, compared to just remit the file(s), turn out to nearly have the same perf footprint then and is therefore not really worth the effort. In case their isn't any useable identifier it could be hidden behind a check for `watch` mode or a opt-in e.g (`options.cache`) with docs that file content changes need a restart when caching && `webpack -w` is used. Albeit assets are less likely to change during one session anyways. I will tackle this after all the defaults-test stuff :)

evilebottnawi:
@michael-ciniawsky i think better to solve this in `this.emitFile`, it is allow cache for all loaders which use `emitFile` and keep loader logic simple

michael-ciniawsky:
Yep, this definitely isn't safe to use

Issue #219, opened 2017-10-29
outputPath is not work when useRelativePath is true


ganxunzou:
- env
```
"file-loader": "^1.1.5",
node:8.8.1
```
- question
```
{ test: /\.(eot|woff|woff2|svg|ttf|png|jpg|jpeg)$/,
        use: [
          {
            loader: 'url-loader',
            options: {
              limit: 100, 
              fallback: 'file-loader', // default
              name: '[name]-[hash:8].[ext]',
              publicPath: 'assets/',
              outputPath: '/imgs/',
              useRelativePath: false, // true : outputPath Failure
            },
          }],
      },
```
Above the code is work OK，but change `useRelativePath:true` , `outputPath` is not work

michael-ciniawsky:
Please include Error Logs or try to describe in more detail **what**  doesn't work. Alternatively provide a test repo for reproduction

qlqllu:
I you see the code, you can see the outputPath is overridden if useRelativePath is true, here is the talking about this:
https://github.com/webpack-contrib/file-loader/pull/222

PR #222, opened 2017-11-01
refactor: function parameter

qlqllu:
* Pass the full filePath into `outputPath` function. Use this parameter, user can generate more dynamic output path
* Do not override 'outputPath' if user have set it

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,12 +19,11 @@ export default function loader(content) {
   });
 
   let outputPath = '';
-  const filePath = this.resourcePath;
 
   if (options.outputPath) {
     // support functions as outputPath to generate them dynamically
     outputPath = (
-      typeof options.outputPath === 'function' ? options.outputPath(url, filePath) : options.outputPath
+      typeof options.outputPath === 'function' ? options.outputPath(url, this.resourcePath) : options.outputPath
     );
   }
 
@@ -32,7 +31,7 @@ export default function loader(content) {
     const issuerContext = (this._module && this._module.issuer
       && this._module.issuer.context) || context;
 
-    const relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join('/');
+    const relativeUrl = issuerContext && path.relative(issuerContext, this.resourcePath).split(path.sep).join('/');
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
 
--- a/src/index.js
+++ b/src/index.js
@@ -36,14 +36,11 @@ export default function loader(content) {
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
 
-    // do not override outputPath if it's set
-    if (!outputPath) {
-      // eslint-disable-next-line no-bitwise
-      if (~relativePath.indexOf('../')) {
-        outputPath = path.posix.join(outputPath, relativePath, url);
-      } else {
-        outputPath = relativePath + url;
-      }
+    // eslint-disable-next-line no-bitwise
+    if (~relativePath.indexOf('../')) {
+      outputPath = path.posix.join(outputPath, relativePath, url);
+    } else {
+      outputPath = relativePath + url;
     }
 
     url = relativePath + url;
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -19,35 +19,35 @@ export default function loader(content) {
   });
 
   let outputPath = '';
+  const filePath = this.resourcePath;
```
michael-ciniawsky: `const filePath` => `const resourcePath`

src/index.js
```diff
@@ -19,35 +19,35 @@ export default function loader(content) {
...
     const relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join('/');
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
-    // eslint-disable-next-line no-bitwise
-    if (~relativePath.indexOf('../')) {
-      outputPath = path.posix.join(outputPath, relativePath, url);
-    } else {
-      outputPath = relativePath + url;
+
+    // do not override outputPath if it's set
```
michael-ciniawsky: Is this a breaking change, can you please provide a before/after example to clearify?

src/index.js
michael-ciniawsky: Isn't `url` (L15) already returning the `filePath`?  What is missing there ?

src/index.js
```diff
@@ -19,35 +19,35 @@ export default function loader(content) {
   });
 
   let outputPath = '';
+  const filePath = this.resourcePath;
 
   if (options.outputPath) {
     // support functions as outputPath to generate them dynamically
     outputPath = (
-      typeof options.outputPath === 'function' ? options.outputPath(url) : options.outputPath
+      typeof options.outputPath === 'function' ? options.outputPath(url, filePath) : options.outputPath
```
michael-ciniawsky: Please give an example for `url` && `filePath`

src/index.js
qlqllu: The `url` is relative path, the filePath is a full absolute path. I need to return outputPath depends on the path info in the full path.

qlqllu: For this one: `const filePath => const resourcePath`, I just move this line from below

src/index.js
qlqllu: This is my config in webpack:
```
{
      test: /\.(png|jpg|gif|svg|eot|ttf|woff|woff2)$/,
      loader: 'url-loader',
      options: {
        limit: 10000,
        outputPath: (rPath, fullPath) => {
          return path.relative(__dirname, fullPath).replace(/\\/g, '/');
        },
        useRelativePath: true,
        name: fullPath => {
          if(fullPath.indexOf('builder') > -1){
            return path.relative(path.join(__dirname, 'builder'), fullPath).replace(/\\/g, '/');
          }else{
            return path.relative(path.join(__dirname, 'stemapp'), fullPath).replace(/\\/g, '/');
          }
        }
      }
    }
```
This is the CSS code:
```
.bg-img{
    background-image: url('webapp.png')
  }
```
In this case, the url is `./webapp.png`, the `filePath` is `d:/root_path/..../webapp.png`

src/index.js
qlqllu: I don't have a before sample, I just use this loader these days.

For the after sample, Please see my webpack config. I want to set both the `name` and the `outputPath`

src/index.js
michael-ciniawsky: Please post your file tree, this looks confusing and I don't really get what you're up to tbh 🙃

qlqllu:
@d3viant0ne I agree with you that I should not include the outputPath override changes in this PR, it's removed now. Can you review?

qlqllu:
For the outputPath override, I found an issue about this: https://github.com/webpack-contrib/file-loader/issues/219

Can I send a PR to fix this one?

src/index.js
qlqllu: My file tree and the case is a little complex, it's like this:
```
builder
  |_some folders
stemapp
  |_widgets
    |_style.css
    |_webapp.png
```

What I want to do is: let the outputPath return a path but the name return a different path.

PR #231, opened 2017-12-01
feat(index): add `dynamicPublicPath` option (`options.dynamicPublicPath`)

anvar:
Not sure if this is of any value for the community, but thought I would share it anyway.

This is to do with utilising a CDN on a separate domain from the main one. In my particular scenario the base URL is fetched from a remote service before loading assets. The base URL is stored in `window` and when resolving asset urls it must be prepended to form an absolute url. `publicPath` almost fits the bill, but it runs the return value through `JSON.stringify`, which makes it unsuitable for my use case.

### DIFF
--- a/test/options/__snapshots__/dynamicPublicPath.test.js.snap
+++ b/test/options/__snapshots__/dynamicPublicPath.test.js.snap
@@ -1,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Options dynamicPublicPath {Function} 1`] = `"throw new Error(\\"Module parse failed: /Users/anvar/projects/quanttech/file-loader/src/index.js??ref--0!/Users/anvar/projects/quanttech/file-loader/test/fixtures/file.png Identifier directly after number (1:23)\\\\nYou may need an appropriate loader to handle this file type.\\\\n| module.exports = test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png;\\");"`;
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ module.exports = {
         use: [
           {
             loader: 'file-loader',
-            options: {}  
+            options: {}
           }
         ]
       }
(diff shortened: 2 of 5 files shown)


test/options/__snapshots__/dynamicPublicPath.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Options dynamicPublicPath {Function} 1`] = `"throw new Error(\\"Module parse failed: /Users/anvar/projects/quanttech/file-loader/src/index.js??ref--0!/Users/anvar/projects/quanttech/file-loader/test/fixtures/file.png Identifier directly after number (1:23)\\\\nYou may need an appropriate loader to handle this file type.\\\\n| module.exports = test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png;\\");"`;
```
michael-ciniawsky: There is something off as the loader throws an `{Error}`

anvar: Thanks for the feedback! Sure thing, the main thing is that `publicPath` takes the return value from the function and wraps it in `JSON. stringify`, which means that you cannot use variables that are to be resolved during runtime in the application. An example is `window.SomeVariable`.

In regards to the test, it is a bit odd. It did the same when I ran it locally, the I ran `npm test -- -u` as suggested by the script after which it worked flawlessly.

michael-ciniawsky:
What about a second argument for `options.publicPath` instead e.g 

```js
{
   loader: 'file-loader',
   options: {
      publicPath (url, { stringify: false }) {
        return ...
   }
}
```

Could you please try to provide an example (repo) how this works in more detail ? I'm not 💯 understanding the usage yet tbh :)

anvar:
Having a second argument for `options.publicPath` would definitely work. I can change it if it is preferable! 

Unfortunately there is not easy way to upload a repo that show-cases this use case as everything is in our corporate repos. I can see whether I can get permission to upload something to Github, but in the mean time I can try to explain the use case a bit better.

Essentially the app consists of two parts; 1) a shell and module loader, and 2) the modules themselves. This change is to support assets in the modules.

The sequence is as follows:
1) The user loads `https://www.foo.com`, which loads the shell.
2) The shell fetches a module `baseUrl` such as `https://static.foo.com` and adds it to `window`.
3) The shell injects `script` tags loading modules using the `baseUrl` e.g. `https://static.foo.com/myModule/moduleVersion/dist.js`.
4) `myModule` tries to load `test.svg` but it needs to correctly load it from `https://static.foo.com/myModule/moduleVersion/test.svg` as opposed to `https://www.foo.com/test.svg`

The difference between `publicPath` and `dynamicPublicPath` is that the `baseUrl` parameter is only known at runtime for the app. So with the example above, the `baseUrl` could be  `https://devstatic.foo.com` in dev, and `https://qastatic.foo.com` in qa etc.

An abbreviated `webpack.config` for a module would look like:
```
const packageConfig = require(paths.appPackageJson);
...
{
  loader: require.resolve('file-loader'),
  exclude: [/\.js$/, /\.html$/, /\.json$/],
  options: {
    name: 'static/media/[name].[hash:8].[ext]',
    dynamicPublicPath: function(url) {
      return `window.moduleBaseUrl + "/${packageConfig.name}/${packageConfig.version}/" + ${JSON.stringify(url)}`;
    }
  },
}
```

The config above produces file-loader asset urls in the format of `window.moduleBaseUrl+"/myModule/0.1.0/static/media/logo.5d5d9eef.svg"`. Using the same function with `publicPath` produces `'window.moduleBaseUrl + "/myModule/0.1.0/" + "static/media/logo.5d5d9eef.svg"'`. Notice the single quotes, which result in asset lookup in the form of `window.moduleBaseUrl/myModule/moduleVersion/test.svg`.

Does that make it at all clearer?

anvar:
Hi,

Just checking that the explanation makes sense?

PR #234, opened 2017-12-05
fix(index): add `this.rootContext`

### LINKED ISSUE #233: [webpack v4.0.0-alpha.0] `this.context` => `this.rootContext` (Breaking Change)
(from [webpack#6064](https://github.com/webpack/webpack/issues/6064))

Running `webpack` with `4.0.0-alpha.0` results in the following error, one per file:

```
ERROR in ./assets/images/icons/search.svg
Module build failed: TypeError: Cannot read property 'context' of undefined
    at Object.loader (/Users/drew/Sites/my-site/node_modules/file-loader/dist/index.js:34:49)
 @ ./styles.sass 6:40863-40906
```

TheLarkInn:
### `Notable Changes`

- Adds `this.rootContext` (`webpack >= v4.0.0`) for compatibility with `webpack` v4.0.0

### `Issues`

- Fixes #233


src/index.js
```diff
@@ -10,7 +10,7 @@ export default function loader(content) {
 
   validateOptions(schema, options, 'File Loader');
 
-  const context = options.context || this.options.context;
+  const context = this.rootContext;
```
michael-ciniawsky: Was the option for a setting custom `context` be removed on purpose? 
Doesn't this work anymore in general now ?

d3viant0ne: Yes, this is breaking for anything that needs is much like we had to do for etwp 3.1

For any loaders & plugins that require this change, make sure the `next` branch has the `peerDependency` set to `4.x only`

michael-ciniawsky: So a `options.context` can't technically exists anymore with the changes made to the underlying logic in webpack (`rootContext`) ?

sokra: `const context = options.context || this.rootContext || this.options && this.options.context`

`this.options` doesn't exist anymore. It was already deprecated in webpack 3. webpack 4 adds `this.rootContext` as replacement for the needed `context` information. We could backport this if it helps.

michael-ciniawsky: Is there particular reason to name it `rootContext` instead of `context` ? :) I assumed the meaning/logic also changed somehow due to the rename and the correlation between `config.context` => `this.context` is cleaner. There is also no e.g `childContext` etc, or is there anything new ?

👍 for a backport of `rootContext` to webpack v3.0.0 so loaders/plugins relying on it don't need to add legacy code checks

@TheLarkInn Could you update the PR with @sokra suggestions so we can release a patch instead of a major for this/for now, so folks can continue trying out more advanced stuff (changes) in the meantime ? 😛 (Fairly depressing to fail at the `file-loader` 🙃...)

michael-ciniawsky:
@TheLarkInn Can I take this over or could you apply these changes 

```diff
- const context = this.rootContext
+ const context = options.context || this.rootContext || this.options && this.options.context
```
and rebase please ?

src/index.js
TheLarkInn: Incoming

d3viant0ne:
Closing this because someone needs to take a git class .....

michael-ciniawsky:
The fix should go into master please, so webpack v3.0.0 && 4.0.0 is supported by the `file-loader 
 =< v1.0.0` for `next` we just use `const context = options.context || this.rootContext` then and remove all the legacy stuff. The fix is easy and I need this for `html-loader, style-loader, css-loader, ETWP` e2e testing 😛

PR #236, opened 2017-12-07
Allow any version of webpack as a peerDependency

lencioni:
Webpack 4 alpha has been released. In preparation for v4,
it seems reasonable to change this to allow *. This should cut
down on maintenance and avoid unnecessarily blocking folks
from updating going forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "webpack-defaults": "^1.6.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


lencioni:
Oh actually it seems that * doesn't allow pre-releases

michael-ciniawsky:
The `file-loader` doesn't currently work with `webpack >= v4.0`, see #234. The change in the current form also basically renders the `peerDependency` useless

PR #242, opened 2018-02-07
feat(index): add a new parameter to `publicPath` (`options.publicPath`)

claudiocantara:
This change makes the publicPath function to return the original path of the resource making possible to build custom or relatives paths.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,7 @@ export default function loader(content) {
 
   validateOptions(schema, options, 'File Loader');
 
-  const context = options.context || this.rootContext || this.options && this.options.context
+  const context = options.context || this.rootContext || (this.options && this.options.context);
 
   let url = loaderUtils.interpolateName(this, options.name, {
     context,
--- a/src/index.js
+++ b/src/index.js
@@ -58,7 +58,7 @@ export default function loader(content) {
   if (options.publicPath !== undefined) {
     // support functions as publicPath to generate them dynamically
     publicPath = JSON.stringify(
-      typeof options.publicPath === 'function' ? options.publicPath(url) : options.publicPath + url,
+      typeof options.publicPath === 'function' ? options.publicPath(url, this._module.issuer.resource) : options.publicPath + url,
     );
   }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "file-loader",
-  "version": "1.1.4",
+  "version": "1.1.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/test/options/__snapshots__/publicPath.test.js.snap
+++ b/test/options/__snapshots__/publicPath.test.js.snap
@@ -1,5 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Options publicPath {Function with ressource param} 1`] = `"module.exports = \\"test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";"`;
+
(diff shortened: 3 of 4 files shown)


claudiocantara:
This is a exemple how to use this funcionality

```javascript
//... other webpack config 
loader: "file-loader",
options: {
name: "[name].[ext]?[hash]",
 outputPath: "img/",
 publicPath: ( file , resource ) => {
  // resource can be  'src/templates/pages/firstPath/secondPath/index.pug';

   const basePath = resource.split('src/').pop();  //get the base of resource
   const compareBaseName = path.basename(basePath, '.pug'); 
   const pagesFullPath = basePath.match(/(?:(?=pages\/)(?:.*)[(?=\/)])/).shift();

   const isRoot = pagesFullPath.match(/[^(?:pages)].+$/);

   if (basePath == 'css/style.scss'){
      return `../${file}`;
   }else {
      if( compareBaseName == "index" ) {
         return !isRoot ? `assets/${file}` : `${path.relative(pagesFullPath,'pages/')}/assets/${file}`;

      }   else {
          return `${path.relative(`${pagesFullPath}/anotherPath`,'pages/')}/assets/${file}`;
              }
      }
  },
 // .. at the end it returns '../../assets/img/my-img.png';
```

PS: using the file-loader parameter 'useRelativePath'  didn't resolve my relative path.

claudiocantara:
Test added

claudiocantara:
Sometimes it is necessary to create relative paths for the generated files, but 'useRelativePath' options, is based on the structure where the file is being processed, but should be based on the new structure generated by the bundle. If my generating structure has paths other than the generated one, there will be path conflicts. With this new change the developer can create the own relative path for the generated files.

Example :
```
project
└─src
│ └─folder
│   └─folder2
│      └─folder3
│        │ fileWhoCallImage.ext
│ └─img
│   │ imageReference.png
└─dist
│ └─assets
│   └─img
│     │ imageReference.png
│ └─folder
│   │ fileWhoCallImage.ext 
```

Using 'useRelativePath' in this example. the relative path generated will be '../../../',  but i need '../', 
with the second parameter at publicPath i can find out my relative path with a custom 
algorithm.

michael-ciniawsky:
Could you try to provide a simpler example then the one you posted in https://github.com/webpack-contrib/file-loader/pull/242#issuecomment-363888530 please ? I'm not really in favor of using `this._module.issuer` within the loader if avoidable and would like to triage alternative implementations first (if available)

claudiocantara:
I created a repository with a project that uses this functionality.
 https://github.com/claudiocantara/file-loader-project-test.git 

there's have two npm script , start and start:error.
start , provides my solutions while start:error is using 'useRelativePath'.

PS: You will not pass the entire object this._module.issuer , only the ressorce (this._module.issuer.ressource)

claudiocantara:
Hello, any news on that ?

src/index.js
```diff
@@ -58,7 +58,7 @@ export default function loader(content) {
   if (options.publicPath !== undefined) {
     // support functions as publicPath to generate them dynamically
     publicPath = JSON.stringify(
-      typeof options.publicPath === 'function' ? options.publicPath(url) : options.publicPath + url,
+      typeof options.publicPath === 'function' ? options.publicPath(url, this._module.issuer.resource) : options.publicPath + url,
```
evilebottnawi: Better send second params `this` and you can inside you code what you need, it is universal solution.

claudiocantara: The proposal is to know what is the path of the resource that called the loader. With this information any developer will get to know the level of the tree of the origin file and adapt the output path according.

michael-ciniawsky:
Closing in favor of #166, see https://github.com/webpack-contrib/file-loader/pull/242#discussion_r190930186

Feel free to reopen if and implement the logic in #166 within this PR

PR #244, opened 2018-02-08
docs(README): add `regExp` option (`options.regExp`)

serut:
Hello,

This PR describes the behavior of `regExp` and how to use it.  
I've also fix :  

- the type of the placeholder `[N]`, hope this doesn't alter too much your table layout.
- regExp is not a query parameter, it's an option. Moreover, passing a regExp as a query param is not a good idea, escaping is way too complicated.

Hope you will like it. :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,11 @@ Emits `file.png` as file in the output directory and returns the public URL
 |:--:|:--:|:-----:|:----------|
 |**`name`**|`{String\|Function}`|`[hash].[ext]`|Configure a custom filename template for your file|
 |**`context`**|`{String}`|`this.options.context`|Configure a custom file context, defaults to `webpack.config.js` [context](https://webpack.js.org/configuration/entry-context/#context)|
-|**`publicPath`**|`{String\|Function}`|[`__webpack_public_path__ `](https://webpack.js.org/api/module-variables/#__webpack_public_path__-webpack-specific-)|Configure a custom `public` path for your files|
-|**`outputPath`**|`{String\|Function}`|`'undefined'`|Configure a custom `output` path for your files|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,7 +107,7 @@ You can configure a custom filename template for your file using the query param
 |**`[name]`**|`{String}`|`file.basename`|The basename of the resource|
 |**`[path]`**|`{String}`|`file.dirname`|The path of the resource relative to the `context`|
 |**`[hash]`**|`{String}`|`md5`|The hash of the content, hashes below for more info|
-|**`[N]`**|`{Number}`|``|The `n-th` match obtained from matching the current file name against the query param `regExp`|
+|**`[N]`**|`{String}`, with `N` `{Number}` >0 |``|The `n-th` match obtained from matching the current file name against the `regExp`|
```
michael-ciniawsky: If it is actually a `{String}` e.g `'4'` instead of a `{Number}`, that should be sufficient for the type column imho. You could add a note in the description that N should be `> 0`

README.md
```diff
@@ -197,6 +198,25 @@ import img from './file.png'
...
+```js
+{
+  loader: 'file-loader',
+  options: {
+    regExp: /\/([a-z0-9]+)\/[a-z0-9]+\.png$/,
+    name: '[1]-[name].[ext]'
+  }  
+}
+```
+
```
michael-ciniawsky:
```
// The interpolated name here ...
```

README.md
serut: No you need to write `[4]`, `[` and `]` decorates the number. And explaining why N is a number > 0 is quite long and already explained on the regExp option. Maybe a link above "`N` `{Number}` >0"?

michael-ciniawsky: kk, just `{String}` then please :)

serut: Oki that's not big deal

evilebottnawi:
@michael-ciniawsky friendly ping

PR #246, opened 2018-02-14
fix(index): don't concat `options.outputPath` and `options.publicPath`

### LINKED ISSUE #160: OutputPath is joined with publicPath?
I have set up file loader for fonts however its not working as I thought it would.
my output path is completely different from my public path so I thought by declaring the locations in parameters this would work. e.g.
```
{
  test: /\.(eot|svg|ttf|woff|woff2)$/,
  loader: 'file-loader?name=[name].[ext]&outputPath=fonts/&publicPath=__fonts/'
}
```
the SRC the css should be using is __fonts/name.ext and the files location should be in /fonts/name.ext

but this is not what is happening.
instead im getting
`src: url(__fonts/fonts/font-name.ttf);`

evilebottnawi:
Fixes #160

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,7 @@ export default function loader(content) {
 
   validateOptions(schema, options, 'File Loader');
 
-  const context = options.context || this.rootContext || this.options && this.options.context
+  const context = options.context || this.rootContext || (this.options && this.options.context);
 
   let url = loaderUtils.interpolateName(this, options.name, {
     context,
@@ -27,9 +27,8 @@ export default function loader(content) {
     );
   }
 
-  const filePath = this.resourcePath;
-
   if (options.useRelativePath) {
+    const filePath = this.resourcePath;
     const issuerContext = (this._module && this._module.issuer
       && this._module.issuer.context) || context;
 
@@ -44,13 +43,8 @@ export default function loader(content) {
     }
 
     url = relativePath + url;
-  } else if (options.outputPath) {
-    // support functions as outputPath to generate them dynamically
-    outputPath = typeof options.outputPath === 'function' ? options.outputPath(url) : options.outputPath + url;
-
-    url = outputPath;
   } else {
-    outputPath = url;
+    outputPath += url;
   }
 
   let publicPath = `__webpack_public_path__ + ${JSON.stringify(url)}`;
--- a/test/options/__snapshots__/outputPath.test.js.snap
+++ b/test/options/__snapshots__/outputPath.test.js.snap
@@ -1,5 +1,55 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Options outputPath {Function} 1`] = `"module.exports = __webpack_public_path__ + \\"test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";"`;
+exports[`Options outputPath {Function} 1`] = `
+Object {
+  "assets": Array [
+    "output_path/9c87cbf3ba33126ffd25ae7f2f6bbafb.png9c87cbf3ba33126ffd25ae7f2f6bbafb.png",
+  ],
+  "source": "module.exports = __webpack_public_path__ + \\"9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";",
+}
+`;
 
-exports[`Options outputPath {String} 1`] = `"module.exports = __webpack_public_path__ + \\"/test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";"`;
+exports[`Options outputPath {Function} with \`publicPath\` {Function} 1`] = `
+Object {
+  "assets": Array [
+    "output_path/9c87cbf3ba33126ffd25ae7f2f6bbafb.png9c87cbf3ba33126ffd25ae7f2f6bbafb.png",
+  ],
+  "source": "module.exports = \\"public_path/9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";",
+}
+`;
+
+exports[`Options outputPath {Function} with \`publicPath\` {String} 1`] = `
+Object {
+  "assets": Array [
+    "output_path/9c87cbf3ba33126ffd25ae7f2f6bbafb.png9c87cbf3ba33126ffd25ae7f2f6bbafb.png",
+  ],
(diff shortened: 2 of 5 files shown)


src/index.js
```diff
@@ -10,7 +10,7 @@ export default function loader(content) {
 
   validateOptions(schema, options, 'File Loader');
 
-  const context = options.context || this.rootContext || this.options && this.options.context
+  const context = options.context || this.rootContext || (this.options && this.options.context);
```
evilebottnawi: Just fix `eslint` problem.

src/index.js
```diff
@@ -27,9 +27,8 @@ export default function loader(content) {
     );
   }
 
-  const filePath = this.resourcePath;
-
   if (options.useRelativePath) {
+    const filePath = this.resourcePath;
```
evilebottnawi: `filePath` need only when we use `options.useRelativePath`

test/options/__snapshots__/outputPath.test.js.snap
```diff
@@ -1,5 +1,55 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Options outputPath {Function} 1`] = `"module.exports = __webpack_public_path__ + \\"test/9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";"`;
+exports[`Options outputPath {Function} 1`] = `
+Object {
+  "assets": Array [
+    "output_path/9c87cbf3ba33126ffd25ae7f2f6bbafb.png9c87cbf3ba33126ffd25ae7f2f6bbafb.png",
+  ],
+  "source": "module.exports = __webpack_public_path__ + \\"9c87cbf3ba33126ffd25ae7f2f6bbafb.png\\";",
```
evilebottnawi: Add to tests `assets` and `source` to ensure outputPath don't concat with publicPath

michael-ciniawsky:
@evilebottnawi Could this break setups currently 'relying' on this behavior in some way ?

evilebottnawi:
@michael-ciniawsky i will add tests to check this

evilebottnawi:
@michael-ciniawsky after investigate it will break build in some case. Should be do this in `@next`?

michael-ciniawsky:
Can you give an example of a breaking scenario ? Yes it likely can break some setups, but it's still a bug nevertheless. Bit gray area here 😛 If avoidable a `patch` would be better...

evilebottnawi:
@michael-ciniawsky many people remove public path from output with difference approaches, It's hard to say who does what exactly breaks down, but i agree this is bug, i can take care about new issue after merge this in master :+1:

michael-ciniawsky:
Released in [`v1.1.7`](https://github.com/webpack-contrib/file-loader/releases/tag/v1.1.7) 🎉

Android3000:
Just an FYI, it might be nice to label this as a potentially breaking change in the release notes, as it looks like you were concerned about it, and I can vouch for this almost immediately causing issues with our build.  We were definitely misunderstanding how outputPath was working when using it as a function, and can see by the intent now that it probably wasn't the "right way" and may need to just change our direction. 

Prior to this change the URL was being set to the result of the outputPath function result.  With this change, the URL is now getting **appended** to the end of the outputPath function result.  This caused issues with our asset output and failures to load them.

michael-ciniawsky:
> Prior to this change the URL was being set to the result of the outputPath function result. With this change, the URL is now getting appended to the end of the outputPath function result. This caused issues with our asset output and failures to load them.

Can you give an example ? This sounds like different problem and might be a (accidentically) mistake made in this PR

Android3000:
@michael-ciniawsky Sure -- again, this is probably just a side effect of us (ab)using the outputPath function capability incorrectly.

Prior to this release, providing an outputPath function would see an argument passed in to it of the current URL of the file.  We were using a name like: `"[path][name]-[hash].[ext]"`, and we'd see an argument like "client/fonts/font-icons/icons-123456789.svg" given to the function.  The return value of that outputPath function would end up being where the file would be emitted -- the function basically acted like a way to transform the file path.  So, for instance, I would use it to just organize some things, e.g.
`if (url.startsWith("client/") return url.replace("client/", "assets/");`

So, instead of my file being output and served up with a URL of `"site.com/client/fonts/font-icons/icons-123456789.svg"`, it would instead be `"site.com/assets/fonts/font-icons/icons-123456789.svg"`.  I was using this to basically help organize the output of some assets for in-house vs. vendor vs. other based on the path prefix I was getting.

With this latest change however (due to the removed "else if" conditional on line 47 of the previous version of the file), the URL gets appended to the result of the output path function (see the change on line 47, `outputPath += url`).  Before, in the "else if" that would handle outputPath being present, the url itself was being replaced by the result of the output path (previous version line 51, `url = outputPath`).

Now, the emitted path ends up being like `"assets/fonts/font-icons/icons-123456789.svgclient/fonts/font-icons/icons-123456789.svg"`, and the re-written URL ends up trying to serve it up at `"site.com/client/fonts/font-icons/icons-123456789.svg"`.

michael-ciniawsky:
That's definitely a regression introduced on L47. Tracking issue is here (https://github.com/webpack-contrib/file-loader/issues/249), thx for reporting and PR welcome and appreciated :). cc @evilebottnawi

chtseac:
Bugfix or not, this is a behaviour change in that it changes the public paths of the generated files. It should have increased the major by 1.

jacekk015:
After update webpack-dev-server failed to serve my font files.
`    {
      test: /\.(eot|woff|woff2|ttf|svg)$/,
      loader: 'file-loader',
      options: {
        name: '[name].[ext]',
        outputPath: 'css/fonts/',
        publicPath: '/'
      }
    },`

evilebottnawi:
@jacekk015 wait exactly failed?

jacekk015:
file_loader

File not found only after update. Nothing else changed.

evilebottnawi:
@jacekk015 you should use `publicPath: 'css/fonts/'`

jacekk015:
@evilebottnawi Thanks! It worked!

olore:
I had same problem as @jacekk015. Here is my diff to fix it from 1.1.6 to 1.1.8:

```
-            'file-loader?name=[name]-[hash].[ext]&publicPath=../../../&outputPath=./assets/'
+            'file-loader?name=[name]-[hash].[ext]&publicPath=../../../assets/&outputPath=./assets/'

```

Thank you for the changelog!

vimalans:
Helpful! Thanks.

PR #250, opened 2018-02-20
fix(index): add `ouputPath` only when `publicPath` is not set

### LINKED ISSUE #249: [v1.1.7] `options.outputPath` {Function} is appended to the original URL
Prior to this release, providing an outputPath function would see an argument passed in to it of the current URL of the file. We were using a name like: "[path][name]-[hash].[ext]", and we'd see an argument like "client/fonts/font-icons/icons-123456789.svg" given to the function. The return value of that outputPath function would end up being where the file would be emitted -- the function basically acted like a way to transform the file path. So, for instance, I would use it to just organize some things, e.g.

```js
if (url.startsWith("client/") return url.replace("client/", "assets/");
```

So, instead of my file being output and served up with a URL of `"site.com/client/fonts/font-icons/icons-123456789.svg"`, it would instead be "site.com/assets/fonts/font-icons/icons-123456789.svg". I was using this to basically help organize the output of some assets for in-house vs. vendor vs. other based on the path prefix I was getting.

With this latest change however (due to the removed "else if" conditional on line 47 of the previous version of the file), the URL gets appended to the result of the output path function (see the change on line 47 of the current version, outputPath += url). Before, in the "else if" that would handle outputPath being present, the url itself was being replaced by the result of the output path (previous version line 51, url = outputPath).

Now, the emitted path ends up being like `"assets/fonts/font-icons/icons-123456789.svgclient/fonts/font-icons/icons-123456789.svg"`, and the re-written URL ends up trying to serve it up at `"site.com/client/fonts/font-icons/icons-123456789.svg"`.

evilebottnawi:
### `Issues`

- Closes #249

### `Notable Changes`

- `publicPath` not set => `__webpack_public_path__ + "output_path/file.ext"`
- `publicPath` set => `__webpack_public_path__ + "public_path/file.ext"`
- fix broken `useRelativePath` and tests


src/index.js
```diff
@@ -12,7 +12,7 @@ export default function loader(content) {
 
   const context = options.context || this.rootContext || (this.options && this.options.context);
 
-  let url = loaderUtils.interpolateName(this, options.name, {
+  const originURL = loaderUtils.interpolateName(this, options.name, {
```
michael-ciniawsky: `originURL` => `name || file`

src/index.js
```diff
@@ -37,23 +39,26 @@ export default function loader(content) {
...
     } else {
-      outputPath = relativePath + url;
+      outputPath = relativePath + originURL;
     }
 
-    url = relativePath + url;
+    url = relativePath + originURL;
   } else {
-    outputPath += url;
+    url = outputPath + originURL;
```
michael-ciniawsky: Please add a 'why' comment here :)

src/index.js
```diff
@@ -37,23 +39,26 @@ export default function loader(content) {
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
     // eslint-disable-next-line no-bitwise
     if (~relativePath.indexOf('../')) {
-      outputPath = path.posix.join(outputPath, relativePath, url);
+      outputPath = path.posix.join(outputPath, relativePath, originURL);
     } else {
-      outputPath = relativePath + url;
+      outputPath = relativePath + originURL;
```
michael-ciniawsky: `path.join()` (In general) ? Things like e.g `{ outputPath: 'dist' }`  => `dist263454765.svg` if one misses the `/` atm

src/index.js
evilebottnawi: @michael-ciniawsky agree, but `interpolateName` return `url` (https://github.com/webpack/loader-utils/blob/master/lib/interpolateName.js#L96), maybe better leave as is (`originURL` || `originalURL`)

michael-ciniawsky:
What about the issue where the result of `{Function}`  gets appended to the path instead of replacing it ?  Assuming this is fixed by https://github.com/webpack-contrib/file-loader/pull/250#discussion-diff-169345588R49 but better double check :)

evilebottnawi:
@michael-ciniawsky can you give example config?

src/index.js
```diff
@@ -12,48 +12,41 @@ export default function loader(content) {
...
-      && this._module.issuer.context) || context;
+    const issuerContext = context || (this._module && this._module.issuer
+      && this._module.issuer.context);
 
     const relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join('/');
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
     // eslint-disable-next-line no-bitwise
     if (~relativePath.indexOf('../')) {
-      outputPath = path.posix.join(outputPath, relativePath, url);
```
michael-ciniawsky: Why does this need to be changed ?

src/index.js
```diff
@@ -12,48 +12,41 @@ export default function loader(content) {
...
+      outputPath = path.join(outputPath, relativePath, originalURL);
     } else {
-      outputPath = relativePath + url;
+      outputPath = path.normalize(relativePath + originalURL);
```
michael-ciniawsky: This still concats 'malformed' `options.outputPath` fragments the wrong way ? `{ outpathPath: 'dist' }` => `dist6254824.svg`  (missing `'/'` (`dist/457438t3z58.svg`)) ?

michael-ciniawsky:
```js
path.normalize(path.join(relativePath, originalURL))
```

src/index.js
```diff
@@ -12,48 +12,41 @@ export default function loader(content) {
 
   const context = options.context || this.rootContext || (this.options && this.options.context);
 
-  let url = loaderUtils.interpolateName(this, options.name, {
+  const originalURL = loaderUtils.interpolateName(this, options.name, {
```
michael-ciniawsky: Since there seems to be no `url` variable anymore, please change it back to `url` then

src/index.js
```diff
@@ -12,48 +12,41 @@ export default function loader(content) {
...
-  let outputPath = '';
-
-  if (options.outputPath) {
-    // support functions as outputPath to generate them dynamically
-    outputPath = (
-      typeof options.outputPath === 'function' ? options.outputPath(url) : options.outputPath
-    );
-  }
+  let outputPath = (
+    typeof options.outputPath === 'function' ? options.outputPath(originalURL) : path.join((options.outputPath || '') + originalURL)
```
michael-ciniawsky:
```js
: path.join(options.outputPath || '', url)
```

Please use a multiline ternary for readability here

```js
 typeof x === 'function'
  ? y
  : z
```

src/index.js
michael-ciniawsky: Maybe already 'fixed' here https://github.com/webpack-contrib/file-loader/pull/250#discussion-diff-169404291R22

src/index.js
evilebottnawi: @michael-ciniawsky because all URL is posix, no need use `posix` here

michael-ciniawsky:
I try to provide an example, just one sec I got distracted

src/index.js
evilebottnawi: @michael-ciniawsky eslint fix, old `webpack-default`

michael-ciniawsky: `eslint-disable`

src/index.js
evilebottnawi: @michael-ciniawsky let's add tests too :+1:

michael-ciniawsky:
`file-loader >= v1.1.7`
```js
{
  loader: 'file-loader',
  options: {
    name: '[name].[ext]',
    outputPath (url) {
      return url
    }
  }
}
```

```
file.svgfile.svg
```

file_loader

evilebottnawi:
@michael-ciniawsky let's fix this

evilebottnawi:
@michael-ciniawsky just add tests, all works as expected here :+1:

src/index.js
```diff
@@ -12,48 +14,45 @@ export default function loader(content) {
...
-  if (options.outputPath) {
-    // support functions as outputPath to generate them dynamically
-    outputPath = (
-      typeof options.outputPath === 'function' ? options.outputPath(url) : options.outputPath
-    );
-  }
+  let outputPath = (
+    typeof options.outputPath === 'function'
+      ? options.outputPath(url)
+      : path.join((options.outputPath ? options.outputPath.replace(/\/?$/, '/') : '') + url)
```
michael-ciniawsky: 🙃 ?

```js
console.log(path.join('dirname', 'file.extname'))
console.log(path.join('dirname/', 'file.extname'))
console.log(path.join('dirname/', '/file.extname'))

// => 'dirname/file.extname'
```
👇 
```js
path.join(outputPath || '', url)
```
?

src/index.js
```diff
@@ -12,48 +14,45 @@ export default function loader(content) {
...
-  let publicPath = `__webpack_public_path__ + ${JSON.stringify(url)}`;
+  let publicPath = null;
 
   if (options.publicPath !== undefined) {
     // support functions as publicPath to generate them dynamically
     publicPath = JSON.stringify(
-      typeof options.publicPath === 'function' ? options.publicPath(url) : options.publicPath + url,
+      typeof options.publicPath === 'function'
+        ? options.publicPath(url)
+        : path.join(options.publicPath.replace(/\/?$/, '/'), url),
```
michael-ciniawsky: See comment for `outputPath`

src/index.js
```diff
@@ -12,48 +14,45 @@ export default function loader(content) {
...
     const relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join('/');
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
     // eslint-disable-next-line no-bitwise
     if (~relativePath.indexOf('../')) {
-      outputPath = path.posix.join(outputPath, relativePath, url);
+      outputPath = path.join(outputPath, relativePath, url);
     } else {
-      outputPath = relativePath + url;
+      outputPath = path.normalize(relativePath + url);
```
michael-ciniawsky: Also `path.join()` should concat && **normalize** (`path.join()` calls `path.normalize()`) this better

src/index.js
```diff
@@ -3,6 +3,8 @@ import loaderUtils from 'loader-utils';
 import validateOptions from 'schema-utils';
 import schema from './options.json';
 
+/* eslint-disable multiline-ternary */
```
michael-ciniawsky: Move this to the top please

evilebottnawi: @michael-ciniawsky before `import`?

src/index.js
evilebottnawi: @michael-ciniawsky need for `outputPath` => `dir` without trailing slash

src/index.js
evilebottnawi: @michael-ciniawsky also see above

src/index.js
evilebottnawi: @michael-ciniawsky should we change all `path.join` to `path.normalize`?

src/index.js
michael-ciniawsky: Yep

src/index.js
evilebottnawi: @michael-ciniawsky done

src/index.js
evilebottnawi: @michael-ciniawsky done :+1:

src/index.js
evilebottnawi: @michael-ciniawsky oh stupid error i use `+`, instead `,`, fixed :smile:

michael-ciniawsky: Why/Where ? :)

evilebottnawi:
@michael-ciniawsky all done :+1:

src/index.js
evilebottnawi: @michael-ciniawsky after `options.publicPath || ''` we don't need this comment :+1:

src/index.js
michael-ciniawsky: But why was it added then ? Could this break anything ? I'm definitely not against removing it, but I don't get why is eventually needed here in the first place

src/index.js
```diff
@@ -12,48 +12,41 @@ export default function loader(content) {
...
     const relativeUrl = issuerContext && path.relative(issuerContext, filePath).split(path.sep).join('/');
 
     const relativePath = relativeUrl && `${path.dirname(relativeUrl)}/`;
     // eslint-disable-next-line no-bitwise
     if (~relativePath.indexOf('../')) {
-      outputPath = path.posix.join(outputPath, relativePath, url);
+      outputPath = path.join(outputPath, relativePath, url);
     } else {
-      outputPath = relativePath + url;
+      outputPath = path.join(relativePath + url);
```
michael-ciniawsky: No `+` :)

src/index.js
evilebottnawi: @michael-ciniawsky 
```js
path.join((options.outputPath ? options.outputPath.replace(/\/?$/, '/') : '') + url); // `+`
```

src/index.js
evilebottnawi: @michael-ciniawsky already fix :smile: Faster then you post comment :smile:

michael-ciniawsky:
The commit message isn't quite correct anymore, any suggestions for a better one ? :)

evilebottnawi:
@michael-ciniawsky no idea, big bug fix? :smile: Feel free to change commit message

michael-ciniawsky:
publicPath set => __webpack_public_path__ + "public_path/file.ext"
publicPath not set => __webpack_public_path__ + "output_path/file.ext"

What happens if both `publicPath && outputPath` options are set ?

evilebottnawi:
@michael-ciniawsky for output we use outputPath + url, for public we use publicPath + url (without concat output path)

michael-ciniawsky:
Could you split them please so I can rebase merge instead ?

````
fix(index): normalize and concat paths via `path.join()`
fix(index): don't append `outputPath` to the original `url` (`options.outputPath` `{Function}`)
fix(index): `context` takes precedence over `issuer.context` (`options.useRelativePath`)
```

evilebottnawi:
@michael-ciniawsky done

PR #253, opened 2018-02-21
fix(index): handle protocol URL's correctly (`options.publicPath`)

### LINKED ISSUE #252: [v1.1.8] publicPath does not work with protocol/absolute URLs
I could be misusing this option but up to/including 1.1.7,
using the **publicPath** for absolute URLs like `publicPath: "https://my.cdn/"` was possible.

In **1.1.8** due to this change:
https://github.com/webpack-contrib/file-loader/commit/26e47ca#diff-1fdf421c05c1140f6d71444ea2b27638R46
(replacing concat with path.join)
it is not possible and the **//** characters are getting replaced with one **/** rendering the absolute URL invalid...

Is there another way to configure CDN / custom absolute URLs?
ie. I want to serve some files from a different protocol/host than the rest of the app...

michael-ciniawsky:
### `Issues`

- Fixes #252

### `Notable Changes`

Since `publicPath` can be a full URL (which can contain protocols), using `path.join()` to concat file paths isn't sufficient here

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,16 +18,28 @@ export default function loader(content) {
     regExp: options.regExp,
   });
 
-  let outputPath = (
-    typeof options.outputPath === 'function' ? options.outputPath(url) : path.join(options.outputPath || '', url)
-  );
+  let outputPath = url;
+
+  if (options.outputPath) {
+    if (typeof options.outputPath === 'function') {
+      outputPath = options.outputPath(url);
+    } else {
+      outputPath = path.join(options.outputPath, url);
+    }
+  }
 
   if (options.useRelativePath) {
     const filePath = this.resourcePath;
-    const issuerContext = context || (this._module && this._module.issuer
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -18,16 +18,30 @@ export default function loader(content) {
     regExp: options.regExp,
   });
 
-  let outputPath = (
-    typeof options.outputPath === 'function' ? options.outputPath(url) : path.join(options.outputPath || '', url)
-  );
+  let outputPath = null;
```
evilebottnawi: `let outputPath = url;` and remove else

src/index.js
```diff
@@ -40,11 +54,16 @@ export default function loader(content) {
 
   let publicPath = null;
```
evilebottnawi: `publicPath = `__webpack_public_path__ + ${JSON.stringify(outputPath)}`;` and remove else

PR #254, opened 2018-02-21
fix(index): use `path.posix` for platform consistency

### LINKED ISSUE #251: [v1.1.8] paths are broken on Windows
Using ElectronJS with latest webpack; after updating to file-loader 1.1.8 fonts no longer load (404).

Going back to 1.1.7 fixes the issue. [This](https://github.com/IAmPicard/StarTrekTimelinesSpreadsheet) is the repo (but it's far form minimal, sorry); let me know what information I can provide to help you narrow it down.

```
{
	test: /\.(eot|svg|ttf|woff|woff2)$/,
	use: [{ loader: 'file-loader?name=font/[name]__[hash:base64:5].[ext]' }],
	include: defaultInclude
}
```

michael-ciniawsky:
### `Issues`

- Fixes #251 

### `Notable Changes`

Use `path.posix` instead of `path` to avoid platform specifics of the `path` module to keep paths consistent

### `Dependants`

- #253  (Ideally lands before this PR)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -24,7 +24,7 @@ export default function loader(content) {
     if (typeof options.outputPath === 'function') {
       outputPath = options.outputPath(url);
     } else {
-      outputPath = path.join(options.outputPath, url);
+      outputPath = path.posix.join(options.outputPath, url);
     }
   }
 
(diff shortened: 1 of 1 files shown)


garygreen:
Is it possible to have some tests for this? It's frustrating to have these kind of regressions on Windows, having tests would of avoided this. I'm a total webpack noob so not sure how you would go about testing it but would be cool if there was! 😄

evilebottnawi:
@garygreen tests already exist when we run CI on window `continuous-integration/appveyor/pr`, but something wrong with CI right now, in near future we fix this

PR #274, opened 2018-04-26
feat: add 'stringifyPublicPath' option

BobNisco:
This adds a new top-level option called `stringifyPublicPath` (set to true by default). By default, this has no discernable difference between the current functionality of file-loader.

When set to false, the value of `publicPath` will not be processed with `JSON.stringify`. This is useful for the cases where the value of `publicPath` is dynamic and driven by runtime variables.

A strong example has been added to the README that outlines the use case for CDN switching at runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ module.exports = {
         use: [
           {
             loader: 'file-loader',
-            options: {}  
+            options: {}
           }
         ]
       }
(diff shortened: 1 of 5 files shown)


evilebottnawi:
@BobNisco Thanks for PR. My first notice - adding new options for `file-loader` is bad, he should be very easy (only in `1.1.11` he is work in 99%, also future major release remove `useRelativePath` options, because normalization urls should do uglificator or minificator).

 Can your describe use case more verbosely? Why you need dynamic this feature? Why you don't use `publicPath` with static value with CDN?

BobNisco:
I was looking for ways to not add a new option but did not see an obvious path forward due to the way that `publicPath` is derived in the current logic. If you have other suggestions, I am open to making modifications here.

Perhaps my explanation in the README isn't verbose enough. In the case where your application has different environments, let's say, staging and production, then simply having 1 static value on CDN is not enough.

Because `file-loader` enforces a static value you would need to build your webpack bundles 2 times: once for your staging CDN endpoint, and once for your production CDN endpoint. This is inefficient.

This PR alleviates that by allowing us to return a slightly more arbitrary value from `file-loader` that allows the file path to be determined at runtime instead of pre-compiled. Now, you only need to build your assets one time and they can be used in many different environments.

BobNisco:
> why don't use difference publicPath value for stag and prod?

This PR accomplishes us being able to use different `publicPath` values for staging and production but utilizing the same static bundle between the two.

> It is very bad practice use window.CDN as in your example, it is insecurity. Static bundle for static, not for dynamic 😄 

The `window.CDN` example is a contrived example to show off functionality at a basic level.

> Also don't forget about other loader, we can handle you value, how they should do this?

Could you elaborate a bit more on this?

evilebottnawi:
@BobNisco 
> This PR accomplishes us being able to use different publicPath values for staging and production but utilizing the same static bundle between the two.
Why don't use `publicPath: isStag ? 'http://stag-cdn.com/' : 'http://prod-cdn.com/' `

> The window.CDN example is a contrived example to show off functionality at a basic level.
README don't contain bad practice, some people can do what you write.

> Could you elaborate a bit more on this?

Other loader can use `file-loader` also after `file-loader` can be other loader, it is rare use case, but it is possible.

evilebottnawi:
Will be great if you provide real use case where you have problem(s) with examples.

BobNisco:
Looks like I had a misunderstanding with some parts of webpack. Here's the solution I actually ended up needing:

1. Define `__webpack_public_path__ ` on the fly. see: https://webpack.js.org/guides/public-path/#on-the-fly
2. Remove extraneous file-loader config for setting public path
3. Defer to `__webpack_public_path__` and `file-loader` defaults to enable runtime URL switch.

Appreciate the review @evilebottnawi. I am going to close this PR out as it is not needed.

Issue #164, opened 2017-06-05
Forwarding Data to other chained loaders


zouhir:
Hi, thanks all for this great loaders. I made [lqip-loader](https://github.com/zouhir/lqip-loader) for webpack and I am curious if `file-loader` ability to forward data (path) to other loaders could be supported. I'm happy to help and would be proud to contribute where I can.

### What I found challenging:

> `webpack.config.js`

```
"loaders": ["lqip-loader", "file-loader"]
```
> `lqip-loader`

inside my loader I get a `string` value from `file-loader` of `"module.exports = __... .jpg"` and to consume that in my loader I have to go through some string manipulation of regex and replacing which I felt could have been done differently.

Happy to know what kind of thoughts you have in order for developers to easily and safely chain after file loader.

Many thanks!

evilebottnawi:
@zouhir Can use provide use case and which `string manipulation` your did? Maybe we can found other solution to this, thanks!

zouhir:
@evilebottnawi thanks for getting back!

When I pipe my loader after `file-loader` I don't get a file path, instead I get something like:

```js
"module.exports = /assets/media/images/sky.jpg"
```

and to use it inside my loader I have to do something like:

- string trim white spaces
- string replace `module.exports` with ``
- string replace `"` with ``
- string replace `=` with ``

and then I'll end up with 

```js
"/assets/media/images/sky.jpg" 
```

which I wanted.

is there a way to pass data from file-loader to other loader where we get the file path without having to do all that string replace? the only reason I didn't keep it is because I thought it's probably not the right way to pass things after file-loader

I'm happy to collaborate further @evilebottnawi please let me know how can I make things clearer.

evilebottnawi:
@zouhir why don't use only `lqip` `"loaders": ["lqip-loader"]`?

lukeed:
I imagine that something like this is what @zouhir wants:

```js
module: {
  rules: [{
    test: /\.(jpg|png)$/,
    loader: 'file-loader!lqip-loader'
  }]
}
```

This way, `file-loader` should return its file paths (or files) so that other chained loaders may work with the same files. _FWIW, I don't know if this is what currently happens, but I would expect this to be the behavior._

I may be wrong/misunderstanding. Just trying to rephrase the question since there seems to be a disconnect here 😬

michael-ciniawsky:
**file-loader**
```diff
- return `export default = ${publicPath};`;
[...]

PR #321, opened 2019-03-05
Exposing output path to subsequent loaders

cduivis:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [x] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

In certain projects you want to parse a lot of files which have their own dependencies. When using file-loader you can tell it into which folder the files need to be written. The problem is the loader processing the files, doesn't know to which paths all the files are written and thus cannot make a correct relative path dependency on those files. By pitching the calculated output path of file-loader to all subsequent loaders they then can use this information to correctly resolve a relative path.

In my case I have multiple .html files which include some .less files specific to it. These files need to be placed in the same folder. I don't want to worry inside the .html file about the setup of webpack so I just want to include the .less file relatively. For this to work correctly extract-loader should know where the .html file needs to be written to create correct relative paths for the .less files.

I've already setup a version of  [extract-loader](https://github.com/NetMatch/extract-loader) which uses these changes to resolve paths relatively.

### Breaking Changes

No breaking changes.

### Additional Info

There can be parts of the path which can't be correctly resolved yet like the `[hash]` as it relies on the content of the file for the hash to be formed. The output path for this file will just keep `[hash]` as part of the output path name. As it's unique to this file it will still resolve relative paths correctly.

### DIFF
--- a/src/cjs.js
+++ b/src/cjs.js
@@ -2,3 +2,6 @@ const loader = require('./index');
 
 module.exports = loader.default;
 module.exports.raw = loader.raw;
+module.exports.pitch = function() {
+  loader.getOutputPath(this);
+};
--- a/src/cjs.js
+++ b/src/cjs.js
@@ -2,6 +2,4 @@ const loader = require('./index');
 
 module.exports = loader.default;
 module.exports.raw = loader.raw;
-module.exports.pitch = function() {
-  loader.getOutputPath(this);
-};
+module.exports.pitch = loader.pitch;
--- a/src/index.js
+++ b/src/index.js
@@ -5,19 +5,23 @@ import validateOptions from 'schema-utils';
 
 import schema from './options.json';
 
-export default function loader(content) {
+function getOptions() {
   const options = loaderUtils.getOptions(this) || {};
 
   validateOptions(schema, options, 'File Loader');
 
-  const context = options.context || this.rootContext;
+  return options;
+}
 
-  const url = loaderUtils.interpolateName(this, options.name, {
+function createUrl(context, options, content) {
+  return loaderUtils.interpolateName(this, options.name, {
     context,
     content,
     regExp: options.regExp,
   });
+}
 
+function createOutputPath(url, options, context) {
   let outputPath = url;
 
   if (options.outputPath) {
@@ -28,6 +32,15 @@ export default function loader(content) {
     }
   }
 
+  return outputPath;
+}
+
+export default function loader(content) {
+  const options = getOptions.call(this);
+  const context = options.context || this.rootContext;
+  const url = createUrl.call(this, context, options, content);
+
+  const outputPath = createOutputPath.call(this, url, options, context);
   let publicPath = `__webpack_public_path__ + ${JSON.stringify(outputPath)}`;
 
   if (options.publicPath) {
@@ -53,3 +66,11 @@ export default function loader(content) {
 }
 
 export const raw = true;
+
+export function getOutputPath(loaderInstance) {
+  const options = getOptions.call(loaderInstance);
+  const context = options.context || loaderInstance.rootContext;
+  const url = createUrl.call(loaderInstance, context, options, null);
+
+  return createOutputPath.call(loaderInstance, url, options);
+}
--- a/src/index.js
+++ b/src/index.js
@@ -67,10 +67,19 @@ export default function loader(content) {
 
 export const raw = true;
 
-export function getOutputPath(loaderInstance) {
-  const options = getOptions.call(loaderInstance);
-  const context = options.context || loaderInstance.rootContext;
-  const url = createUrl.call(loaderInstance, context, options, null);
+export function pitch() {
+  const options = getOptions.call(this);
+  const context = options.context || this.rootContext;
+  const url = createUrl.call(this, context, options, null);
 
-  return createOutputPath.call(loaderInstance, url, options);
+  const outputPath = createOutputPath.call(this, url, options, context);
+
+  for (let i = this.loaderIndex; i < this.loaders.length; i++) {
+    const loadr = this.loaders[i];
+    const data = loadr.data || {};
+
+    data.outputPath = outputPath;
+
+    loadr.data = data;
+  }
 }
--- a/test/helpers/compiler.js
+++ b/test/helpers/compiler.js
@@ -28,11 +28,13 @@ const module = (config) => {
 
 const plugins = (config) => [].concat(config.plugins || []);
 
+const outputFolder = 'outputs/';
+
 const output = (config) => {
   return {
     path: path.resolve(
       __dirname,
-      `../outputs/${config.output ? config.output : ''}`
+      `../${outputFolder}${config.output ? config.output : ''}`
     ),
     filename: '[name].bundle.js',
   };
@@ -58,12 +60,16 @@ export default function(fixture, config, options) {
   if (options.output) del.sync(config.output.path);
 
   const compiler = webpack(config);
+  const mfs = new MemoryFS();
 
-  if (!options.output) compiler.outputFileSystem = new MemoryFS();
+  if (!options.output) compiler.outputFileSystem = mfs;
 
   return new Promise((resolve, reject) =>
     compiler.run((err, stats) => {
       if (err) reject(err);
+
+      stats.usedFileSystem = mfs;
+      stats.outputFolder = outputFolder;
 
       resolve(stats);
     })
--- a/test/helpers/pitch-outputPath-loader.js
+++ b/test/helpers/pitch-outputPath-loader.js
@@ -1,0 +1,3 @@
+module.exports = function loader() {
+  return this.data.outputPath;
+};
--- a/test/pitch-outputPath.test.js
+++ b/test/pitch-outputPath.test.js
@@ -1,0 +1,66 @@
+import path from 'path';
+
+import webpack from './helpers/compiler';
+
+describe('pitch output path', () => {
+  it('value should be the same as the eventual location its written to', async () => {
+    const config = {
+      rules: [
+        {
+          test: /\.(js)$/i,
+          use: [
+            {
+              loader: path.resolve(__dirname, '../src/cjs'),
+            },
+            {
+              loader: path.resolve(
+                __dirname,
+                './helpers/pitch-outputPath-loader'
+              ),
+            },
+          ],
+        },
+      ],
+    };
+
+    const stats = await webpack('fixture.js', config);
+    const [module] = stats.toJson().modules;
+    const { assets } = module;
+    const filePath = path.resolve(__dirname, stats.outputFolder, assets[0]);
+    const content = stats.usedFileSystem.readFileSync(filePath).toString();
+
+    expect(content).toEqual('[hash].js');
+  });
+
+  it('should handle [path][name].[ext] correctly when using nested folders', async () => {
+    const config = {
+      rules: [
+        {
+          test: /\.(js)$/i,
+          use: [
+            {
+              loader: path.resolve(__dirname, '../src/cjs'),
+              options: {
+                name: '[path][name].[ext]',
+              },
+            },
+            {
+              loader: path.resolve(
+                __dirname,
+                './helpers/pitch-outputPath-loader'
+              ),
+            },
+          ],
+        },
+      ],
+    };
+
+    const stats = await webpack('nested/fixture.js', config);
+    const [module] = stats.toJson().modules;
+    const { assets } = module;
+    const filePath = path.resolve(__dirname, stats.outputFolder, assets[0]);
+    const content = stats.usedFileSystem.readFileSync(filePath).toString();
+
+    expect(content).toEqual('nested/fixture.js');
+  });
+});


cduivis:
@evilebottnawi a review from anyone of the team would be appreciated.

cduivis:
Because of the strict building rules and committing rules that wasn't really an option.
Firstly, I need to adhere to the commit rules your repository has configured.

Secondly, your automated PR bot required me to fix the security vulnerabilities that NPM reported before it would accept the PR.

So what would you suggest?

evilebottnawi:
@cduivis avoid this, it is chore job, feel free if you CI was failed, before merge i fix it and rebase you branch

cduivis:
So if I understand correctly you want me to undo the audit changes, and then you will re-fix them on your end?

evilebottnawi:
@cduivis yes, you are right

cduivis:
Ok I've removed the audit changes as requested ;)

evilebottnawi:
Introduce pitch is breaking change, just create util directory with utils and export their as part of laoder api

cduivis:
Could you explain to me how or why this is a breaking change?
The file loader still does exactly the same as it did before it only pitches data to subsequent loaders, which can be useful for them.
This means if you have file-loader followed by css-loader. That css-loader can now access the output directory that file-loader is going to use to write its files to.
Also non of the unit tests are breaking and they have stayed the same.

It's sending a variable which can be used in other loaders. I don't see how an util directory would help with that as nothing has been written to disk yet?

evilebottnawi:
Here right solution https://github.com/webpack-contrib/file-loader/issues/164#issuecomment-312772005

cduivis:
No it isn't that change is trying to expose the data to the next loader after file-loader in the "normal" flow. What this PR tries to do is expose that data to loaders which are executed before the file-loader in the normal flow. We do this by making use of the pitching phase to make sure that data can already be accessed.

But I'm really starting to get the feeling that you didn't even read this PR's Motivation / Use Case.

evilebottnawi:
Sorry we can't use this changes, because it is break loader in other cases, we should search other way to provide this ifnromation

evilebottnawi:
Please provide minimum reproducible test repo with problem and describe what you try to solve in README, thanks

cduivis:
You said that before, but as I said can you give me some use cases when it breaks ?
Because it passes al the unit tests, so what are those breaking changes. If we don't know we can't work on a solution.

evilebottnawi:
@cduivis unit tests !== integration tests, you solution looks very harmful, so i am asking you create example to see what you try to solve

cduivis:
> Please provide minimum reproducible test repo with problem
Unit tests have been provided for this new behavior.

> and describe what you try to solve in README, thanks
The intro text of the PR describes this as asked in your code of conduct.
In there I've placed a link to an repo of extract-loader which is making use of this change
There are unit tests in that project as well which you can look into / execute to see what it does.

But I can update the README, which explains how this change could be used in loaders other people make.

cduivis:
> @cduivis unit tests !== integration tests, you solution looks very harmful, so i am asking you create example to see what you try to solve

Could you then provide me those integration tests ? Because looking harmfull !== being harmfull.

evilebottnawi:
> The intro text of the PR describes this as asked in your contributing guidelines.
In there I've placed a link to an repo of extract-loader which is making use of this change
There are unit tests in that project as well which you can look into / execute to see what it does.

> But I can update the README, which explains how this change could be used in loaders other people make.

Yes please do this

> Could you then provide me those integration tests ? Because looking harmfull !== being harmfull.

Unfortunately no, it is very difficult setup and not for all developers.

After investigate your solution i will have discussion with other developers how to solve this better way

rjgotten:
> @evilebottnawi
> Sorry we can't use this changes, because it is break loader in other cases, we should search other way to provide this information

I'm going to reiterate what @cduivis asked:
which cases?


Because the changes made here should _not_ change pre-existing behavior.

It _adds_ a pitching phase which publishes the desired output URL to all other loaders _further_ into the loader stack, as opposed to the normal loader direction. (You develop JS, so I assume you'll atleast have some passing familiarity with capturing and bubbling events in the browser DOM? It's like that...)
It places this URL onto the data member specifically meant for transferring data between loaders, as per Webpack's published API documentation.

The changes do not touch the regular loader path. I.e.  they do not alter existing behavior. The _ONLY_ danger of an incompatibility here, is the possibility of a name collision with another loader that publishes something in the pitching phase under the same property name.

If that's your gripe; a configuration option can be added to make the pitch behavior opt-in and prevent problems on existing installations.

And yes; some code is being shifted around. I.e. options parsing and URL creation is factored out into re-usable functions, as both the regular loader path as well as the pitching loader path make use of those.
Doesn't mean it changes behavior...

evilebottnawi:
Why this PR invalid:
1. Example of problem not provided
2. PR contains invalid changes like refactor, change version of package and etc (we use aromatically releases)
3. Invalid usage `pitch`

rjgotten:
> Example of problem not provided

There's already a description at the top of this PR.
I assume you want an actual working example to pull down from some repository for demonstration purposes now?

> PR contains invalid changes like refactor, change version of package and etc (we use aromatically releases)

The bulk of which were already covered by backing out those changes. A minimum of changes cannot be removed, as your pre-commit hooks and validation with husky were _broken_ and prevented commits to be succesfully completed. @cduivis will have more details on that.

> Invalid usage `pitch`

Please explain what's invalid about it, then?
Because as far as I can tell; it's not.

Also; we've been using this succesfully from our own fork on internal projects as well. No issues. So where's the fire? Maybe your integration tests are just flawed? Testing boundaries set more strict than the system requires (and/or the public API dictates) maybe? 

Couldn't tell though as you're refusing to publish those.

evilebottnawi:
> I assume you want an actual working example to pull down from some repository for demonstration purposes now?

Yes

> The bulk of which were already covered by backing out those changes. A minimum of changes cannot be removed, as your pre-commit hooks and validation with husky were broken and prevented commits to be succesfully pushed up. @cduivis will have more details on that.

Just do rebase and don't touch unnecessary, if you have problem with something please open new issue and it will be fixed in other PR 

> Please explain what's invalid about it, then?
> Because as far as I can tell; it's not.

> ...

As I said above, there are other possibilities of use and they are not covered, I can publish examples later, now I do not have time for this, that is why I ask you to give an example

cduivis:
> Just do rebase and don't touch unnecessary, if you have problem with something please open new issue and it will be fixed in other PR

The rebase is done.

> Example of problem is not provided.

If you want to see this change in action please pull down the following repository and run its unit tests [extract-loader](https://github.com/NetMatch/extract-loader)
If you just want to look into the code what it does:
- [here is the wepack setup for nested files](https://github.com/NetMatch/extract-loader/blob/master/test/support/compileNested.js)
- [here you can see the unit tests that test if the correct relative paths are generated](https://github.com/NetMatch/extract-loader/blob/master/test/extractLoader.relativePath.test.js)

> As I said above, there are other possibilities of use and they are not covered, I can publish examples later, now I do not have time for this, that is why I ask you to give an example

Please do so, as this back and forth communication is getting us nowhere without well grounded claims.
Hopefully they will finally lead us to a mutual understanding.

evilebottnawi:
Thanks for link and feedback, file-loader in my todo

NormalGaussian:
I could really do with this - is it going to make it in?

anikethsaha:
I think `pitch` metho passes data from `pitch` phase -> `normal` phase of the same loader !!!

`this.callback(.. , .. , .. , data)` seems the accurate approach

rjgotten:
> 
> 
> I think `pitch` metho passes data from `pitch` phase -> `normal` phase of the same loader !!!
> 
> `this.callback(.. , .. , .. , data)` seems the accurate approach

A loader has access to [`this.data`](https://webpack.js.org/api/loaders/#thisdata) which is indeed just a plain object which can hold arbitrary data to share between that loader's pitch and loading phase.

However, any loader has access to the entire _chain_ of loaders via the [`this.loaders`](https://webpack.js.org/api/loaders/#thisloaders) property and also has access to its own index into that array via [`this.loaderIndex`](https://webpack.js.org/api/loaders/#thisloaderindex).

So, it's completely possible to _walk the indices of that array_ starting from the current loader to the end of the array, to populate _earlier_ loaders -- later inthe pitching flow == earlier in the perspective  normal flow -- to pre-populate _their_ `data`.

`this.callback` does something _**entirely**_ different when called from the pitching phase: it shorts out the loader chain, with an early return.
Example: Given a normal loader flow is A->B->C->D, then the pitching phase executes as D->C-B->A. If C uses `this.callback` in its pitching phase, then the entire loader chain halts there and returns whatever C returned from the pitching phase, not visiting B and A in the pitching phase and skipping the normal flow A->B->C->D _completely_.

anikethsaha:
> `this.callback` does something entirely different when called from the pitching phase: it shorts out the loader chain, with an early return.

I meant `this.callback` from normal phase !!

> A loader has access to this.data which is indeed just a plain object which can hold arbitrary data to share between that loader's pitch and loading phase.
>
> However, any loader has access to the entire chain of loaders via the this.loaders property and also has access to its own index into that array via this.loaderIndex.
> 
> So, it's completely possible to walk the indices of that array starting from the current loader to the > end of the array, to populate earlier loaders -- later inthe pitching flow == earlier in the > perspective normal flow -- to pre-populate their data.

Thanks for this explanation 👍 


> So, it's completely possible to walk the indices of that array starting from the current loader to the end of the array, to populate earlier loaders -- later inthe pitching flow == earlier in the perspective normal flow -- to pre-populate their data.

Isnt this a hacky workaround ?

rjgotten:
> Isnt this a hacky workaround ?

Yes and no.

From a purely technical implementation perspective it's just accessing publicly reachable `data` state for another loader. Another loader that hasn't even been reached in the pitching phase yet, so it can't even be running any of its own logic on that `data` to possibly lead to a race condition.
So it's _perfectly_ safe, short of the small odds of there being a naming collision when using string-based properties. There are certainly ways to avoid that as well, by using a `Symbol`.

But conceptually; yes, it's not very nice.
But it's a necessity to cover the blind spot in Webpack's architecture, which only supports bubbling results _up_ in the normal loader flow and does not have an in-built 100% paved cow-path to push metadata _down_ in the pitching flow.

anikethsaha:
> You're suggesting to use `this.callback` -- which does something _**entirely**_ different. When called from the pitching phase: it shorts out the entire loader chain with an early return.
> 
> Here's an example: Given a normal loader flow of (in)->A->B->C->D->(out), then the pitching phase executes as D->C->B->A. If C uses `this.callback` in its pitching phase, then the entire loader chain halts there and returns whatever C returned from the pitching phase, not visiting B and A in the pitching phase and skipping the normal flow A->B->C->D _completely_.
> 
> To put that in ASCII diagrams:
> 
> ```
> Pitch  Normal
> 
> (req)  (resp)
>   |      |
>   v      ^
>   |      |
>   D      D
>   |      |
>   v      ^
>   |      |
>   C      C
>   |      |
>   v      ^
>   |      |
>   B      B
>   |      |
>   v      ^
>   |      |
>   A      A
>    \    /
>    (file)
> ```
> 
> vs
> 
> ```
> Pitch  Normal
> 
> (req)  (resp)
>   |      |
>   v      ^
>   |      |
>   D      |
>   |     / 
>   v    /  
>   |   /   
>   C->- this.callback()
> ```


I meant callback from normal phase !!

I would be something like this 

```
(req)  (resp)
  |        |
  v        ^
  |        |
  D        D
  |        | 
  v        ^
  |        |   
 C       C->- this.callback(.., .., .., data)
 ```

And this is the expected behavior if we simply want to send it to the next chained loader

rjgotten:
> I meant callback from normal phase !!
> 
> I would be something like this
> 
> ```
> (req)  (resp)
>   |        |
>   v        ^
>   |        |
>   D        D
>   |        | 
>   v        ^
>   |        |   
>  C       C->- this.callback(.., .., .., data)
> ```
> 
> And this is the expected behavior if we simply want to send it to the next chained loader

Using `this.callback(.., .., .., data)` from `C` in the normal phase only has `D` left to pass through.
We're talking about data that `B` or `A` need to know from `C`, so `this.callback(.., .., .., data)` from `C` in the normal phase is too late.

anikethsaha:
> > I meant callback from normal phase !!
> > I would be something like this
> > ```
> > (req)  (resp)
> >   |        |
> >   v        ^
> >   |        |
> >   D        D
> >   |        | 
> >   v        ^
> >   |        |   
> >  C       C->- this.callback(.., .., .., data)
> > ```
> > 
> > 
> > And this is the expected behavior if we simply want to send it to the next chained loader
> 
> Using `this.callback(.., .., .., data)` from `C` in the normal phase only has `D` left to pass through. Here, we're talking about data that either `B` or `A` needs to know from `C`, so using `this.callback(.., .., .., data)` from `C` in the normal phase is simply _too late_.

Oh, I thought it was just to pass the data to next loader. 
This approach through `this.data` seems fine an unique to file loader and other loaders which moves or writes new file. 

@rjgotten thanks a lot for clearing this.

Yankovsky:
Any news on this issue?

Yankovsky:
As I see this issue, there are two ways to solve it:
1. Add pitch phase as described by @cduivis and @rjgotten. This approach will make data about path available to all loaders in chain.
2. Use callback https://github.com/webpack-contrib/file-loader/issues/164#issuecomment-312772005 This approach will only provide data about path to next loaders in chain.

Second option is easier to implement, but doesn't cover all possible scenarios. First option has already been implemented in this pr and has only one possible drawback of name collision which could be solved by using Symbol instead of regular string for storing path data.

Yankovsky:
@evilebottnawi Is there anything unclear about purpose or implementation of this pr? I can help with fixes.

evilebottnawi:
No new feature for `file-loader`, because we deprecate this loader after webpack@5 stable release in favor https://webpack.js.org/guides/asset-modules/, we will update docs how to achieve same goal - `runtime` and `sources` changes, sorry for delay, right now you can copy/paste logic from `file-loader`, here is no complex logic and we will not change it, sorry for delay

Why it was not merged? Because using `pitch` can break other loader what rely on this loader. Writing information in loader context is not safe, because it should be immutable, otherwise it will break cache.

rjgotten:
And when exactly will Webpack 5 be stable, i.e. out of pre-release, and have these asset modules marked as a stable feature that is no longer behind the `experiments: true` flag? I.e. when will it actually be production-ready for use, with the kinks worked out?

evilebottnawi:
@rjgotten 

Where I say it? I don't understand why you to talk with me with such a tone?

> And when exactly will Webpack 5 be stable, i.e. out of pre-release, and have these asset modules marked as a stable feature that is no longer behind the experiments: true flag? I.e. when will it actually be production-ready for use, with the kinks worked out?

You can already use webpack@5, migration between webpack@4 and webpack@5 will not be complex or hard, most of loaders and plugins already have compatibility and don't have problems.

> nice to know you have no interest in maintaining this feature on a Webpack 4 branch

Do you read my answer? https://github.com/webpack-contrib/file-loader/pull/321#issuecomment-676414584

Why it was not merged? Because using pitch can break other loader what rely on this loader. 
Writing information in loader context is not safe, because **it should be immutable**, otherwise **it will break cache**.

> I assume that means the entire set of loaders will be end-of-life moving forward to Webpack 5? I.e. no future support for those people and organizations choosing to stay on 4?

Because it has more features in webpack, if you want to stay on webpack@4 you can use this loader all time.

PR #327, opened 2019-04-23
Allow publicPath handler to return a function instead of string

prontiol:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case
Sometimes you might need to construct the path to your assets in runtime (for example when using multi CDN for different types of assets). Currently the only way to change the base path is to update `__webpack_public_path__` and it will affect all the assets at once 😞

If only we could have a function instead of plain string which will be executed in runtime and return the path based on the asset type, variables, cookies or whatever you want to figure out the right base path... oh wait, but we can!
This PR is to rescue!  🎉

### Breaking Changes

No breaking changes.
Current implementation just produces nothing If you return a function  instead of string out of the `options.publicPath` handler because ```JSON.stringify(function) === undefined```
This PR enhances `file-loader` and now you can return a function which will be executed in runtime so you can enjoy all your variables and helper functions in it and construct a publicPath of your dream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,12 @@ module.exports = {
                 if (/images/.test(context)) {
                   return `image_output_path/${url}`;
                 }
+                
+                if (/runtime/.test(context)) {
+                  return function(url) {
+                    return window.customPublicPath + url;
+                  }
+                }
 
                 return `public_path/${url}`;
               },
--- a/src/index.js
+++ b/src/index.js
@@ -41,7 +41,10 @@ export default function loader(content) {
       }${url}`;
     }
 
-    publicPath = JSON.stringify(publicPath);
+    publicPath =
+      typeof publicPath === 'function'
+        ? `(${publicPath})("${url}")`
+        : JSON.stringify(publicPath);
   }
 
   if (typeof options.emitFile === 'undefined' || options.emitFile) {
--- a/test/__snapshots__/publicPath-option.test.js.snap
+++ b/test/__snapshots__/publicPath-option.test.js.snap
@@ -27,6 +27,18 @@ Object {
 }
 `;
 
+exports[`when applied with \`publicPath\` option matches snapshot for \`{Function}\` value returning \`{Function}\` 1`] = `
+Object {
+  "assets": Array [
+    "9c87cbf3ba33126ffd25ae7f2f6bbafb.png",
+  ],
+  "source": "module.exports = (function publicPath(url) {
+              // eslint-disable-next-line no-undef
+              return window.imageCdnBasePath + url;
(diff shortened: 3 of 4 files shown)


prontiol:
@evilebottnawi Can you elaborate the question?

evilebottnawi:
> how you want to solve problem for other loader?

Because it is fix only logic for `file-loader`, but other loader also can emit assets, i think we need solve this on webpack side, so i request minimum reproducible test repo

prontiol:
Well, I am not sure this can be solved on webpack level.
At least this is not something I am willing to fix.

evilebottnawi:
We don't want add new option to file-loader, because in webpack@5 it was built-in loader, so we need solve this problem on webpack level to avoid same problems in future for other developers

prontiol:
Okay no problem, we will just fork it.

evilebottnawi:
@prontiol what is hard to help other developers solve same problem? It is hard to provide minimum reproducible test repo, maybe we can same logic without modify source code

prontiol:
It's not hard so I made this PR. It fixes the problem I encountered and can help others.
You said you didn't want to add new options to `file-loader` because of webpack 5 so I assume this PR won't be merged, that's why we will fork it because we need this fixed now and we can't wait for webpack 5. And I am not sure webpack 5 can do this either.
Generally I didn't expect this resistance as this PR doesn't introduce any breaking changes.
It just handles _function_ output of publicPath in a right way.

Regarding the minimum reproducible repo: test case included in the PR is pretty much self-explanatory.

evilebottnawi:
Stop be toxic. I am open to feedback, unlike how you build a dialogue.

> You said you didn't want to add new options to file-loader because of webpack 5 so I assume this PR won't be merged, that's why we will fork it because we need this fixed now and we can't wait for webpack 5. 

I said that I would not like to add, but if we do not find other solutions, it will be merged

> that's why we will fork it because we need this fixed now and we can't wait for webpack 5

Yes, but we need **feedback** from developers to solve problems and design right architecture to avoid problems in future (for loader and for webpack@5).

> Generally I didn't expect this resistance as this PR doesn't introduce any breaking changes.
It just handles function output of publicPath in a right way.

This does not mean that we should merge all PRs that are sent to us just because they are not breaking change

> Regarding the minimum reproducible repo: test case included in the PR is pretty much self-explanatory.

No, maybe you use invalid approach to storage or organization assets, maybe you need other setup for cdn, and many many cases in other way

What i want? I want example where standard functionality not solve you problem with **real life** example. This will allow us to think how best to solve such problems now and in the future.

prontiol:
I am no longer interested in contributing to this repo.

PR #334, opened 2019-06-12
Dynamic public path

fabb:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

We use the `file-loader` in a [next.js plugin](https://gist.github.com/fabb/cdaad150e79a1dc78aa97fc315d40cb7) to automatically generate asset urls by importing the assets. The asset file names will be hashed, which is good for caching indefinitely. The plugin supports next.js' `assetPrefix` which allows to provide a different cache server host url for static assets. Next.js [uses `__webpack_public_path__` by default](https://github.com/zeit/next.js/blob/7e7f2c0a6df2cf6a48fa4b34beb1d5befe13fa54/packages/next/client/index.js#L47).

The problem is that in our case the `assetPrefix` is dynamic at runtime, and will be set based on environment variables which are **not present at compile time**. The plugin-static gets used during compile time though, and will only use the `assetPrefix` that is based on environment variables present at compile time. [The proper way around this is to make use of `__webpack_public_path__`](https://webpack.js.org/guides/public-path/#on-the-fly). The problem with `file-loader` is that when setting the `publicPath`, it is [not possible to use the `__webpack_public_path__`](https://github.com/webpack-contrib/file-loader/blob/011cc38d4ed470729ff0fbc6c1d7146896603598/src/index.js#L44). Therefore a new setting is needed which allows to prefix with `__webpack_public_path__`.

### Breaking Changes

No breaking changes.


fabb:
That‘s too bad, and I don‘t completely understand it. Could it be that there is a misunderstanding on what the PR does?

1. With vanilla file-loader, when not setting the `publicPath`, it [will default to `__webpack_public_path__ + ${JSON.stringify(outputPath)}`](https://github.com/webpack-contrib/file-loader/blob/master/src/index.js#L31), which is basically the same as what‘s possible with this PR.
2. In this PR, `__webpack_public_path__` is not **set**, it is just returned, so it will be used at runtime to build the public path. Exactly this is necessary so clients can set it according to [the documentation you mentioned](https://webpack.js.org/guides/public-path/#on-the-fly), and get a path depending on runtime environment variables.
3. The change of this PR is purely additional. If `prefixPublicPathWithWebpackPublicPath` is not explicitly set, clients will get the old behavior.

evilebottnawi:
hm, why don't use `publicPath` as `Function`?

fabb:
That is not possible, because the function is executed at buildtime, but I need to modify the publicPath at runtime depending on runtime environment variables that cannot be present at buildtime.

evilebottnawi:
Please provide plugin/app/reproducible repo with problematic and describe what you want to solve in README.md. 

As i said above webpack@5 will have built-in file-loader so you solution can't work in next webpack version, we should search way how to solve this in other way.

evilebottnawi:
oh, looks i see your problem, no need link

fabb:
Ok thx

evilebottnawi:
Let's keep open

evilebottnawi:
I think we need do breaking change again to solve this, `publicPath` option is broken:
1. https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L35 we don't use `JSON.stringify` here, so potentially break code
2. https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L37 we don't add `__webpack_public_path__` here (i think we should this do)

Good catch, anyway `publicPath` should works as workaround for you:
```
publicPath(url) {
  return '__webpack_public_path__' + url;
}
```

fabb:
Thanks for taking the time to understand the problem!

Until I can switch to webpack 5 it‘s fine for me to use my forked package.

evilebottnawi:
I think we should correct this behavior and migrate all tests to webpack@5 to ensure all works es expected

fabb:
> I think we need do breaking change again to solve this, publicPath option is broken:
> 1. ```
> publicPath = options.publicPath(url, this.resourcePath, context); 
> ```
> we don't use JSON.stringify here, so potentially break code.

Not true because there is not `return`. In [line 44](https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L44) it is stringified.

> 2. 
> ```
> publicPath = `${ 
> ```
> we don't add __webpack_public_path__ here (i think we should do this)

Also not true because there is no `return`. [Line 46](https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L46) will be executed in either way.

> Good catch, anyway publicPath should works as workaround for you:
> 
> ```
> publicPath(url) {
>   return '__webpack_public_path__' + url;
> }
> ```

No, this does not work, because of the [`JSON.stringify` in line 44](https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L44). This will add quote around the whole string like this: `"__webpack_public_path__ + passed_url"` which cannot work.

evilebottnawi:
> No, this does not work, because of the JSON.stringify in line 44. This will add quote around the whole string like this: "__webpack_public_path__ + passed_url" which cannot work.

Do you try this? We don't add anything when `publicPath` is function https://github.com/webpack-contrib/file-loader/blob/903aa7544f7c25d524e25a7dc171e9d4d6f80980/src/index.js#L34

fabb:
> Do you try this? We don't add anything when publicPath is function

Not true. Line 44 is also executed in this case afterwards, because there is no `return` in line 35. And yes, I tried it 😉

evilebottnawi:
Oh, yes sorry, anyway you catch a bug and we fix it in next release (wrkaround: use forked repo right now)

fabb:
Sorry, no better ideas for naming 🙈

String|Function is a good idea, i guess you mean for the `publicPath` option? Then we don‘t need to find a better name for `prefixPublicPathWithWebpackPublicPath`, because we don‘t need it at all, right?

evilebottnawi:
Maybe just `postTransformPublicPath` because you can add any own global variable before publicPath not only `__webpack_public_path__` when you use `Function`?

fabb:
Ah, good idea. I just saw that `publicPath` can already be a function, but the result still gets stringified, so for backwards compatibility, a new option is needed anyways (or overload the return value of the `publicPath` function, but that gets too complicated.
Ok will change the PR soon.

evilebottnawi:
Feel free to ping me

fabb:
@evilebottnawi I've implemented the change. I fear it might be a bit confusing for users on how to use the option correctly. Please take a look at the unit tests and tell me what you think.

test/publicPath-option.test.js
```diff
@@ -164,87 +162,140 @@ describe('when applied with `publicPath` and `prefixPublicPathWithWebpackPublicP
...
 
-  // notice that this case will produce invalid urls if __webpack_public_path__ is set to an absolute url
-  it('matches snapshot for `{String}` value as URL', async () => {
-    const config = {
-      loader: {
-        test: /(png|jpg|svg)/,
-        options: {
-          publicPath: 'https://cdn.com/',
-          prefixPublicPathWithWebpackPublicPath: true,
+describe('when applied with `publicPath` and `postTransformPublicPath` option', () => {
```
evilebottnawi: Let's move this tests (all related to `postTransformPublicPath`) to `postTransformPublicPath-option.text.js`

fabb:
ok done! any additional remarks? should we add `resourcePath` and `context` as parameters as well for flexibility?

evilebottnawi:
@fabb let's rebase

fabb:
done!

evilebottnawi:
Good work!

README.md
```diff
@@ -269,6 +269,32 @@ module.exports = {
...
+        options: {
+          publicPath: '/some/path/',
+          postTransformPublicPath: (p) => `__webpack_public_path__ + ${p}`,
+        },
+      },
+    ],
+  },
+};
+```
+
```
evilebottnawi: Oh, i forgot, let's add Example use case for this

fabb: i pushed an example, is this what you mean?

evilebottnawi: oh, i see, just think about additional section in `Example` in README

fabb: ups, i forgot to push the commit i was talking about 😅
i meant this one: https://github.com/webpack-contrib/file-loader/pull/334/commits/460127db553624d14b6af3664986fd41d02082e0

README.md
```diff
@@ -603,6 +603,72 @@ Result:
 path/to/file.png?e43b20c069c4a01867c31e98cbce33c9
 ```
 
+---
```
evilebottnawi: Let's use heading here instead `---`, something like `### Dynamic public path with __webpack_public_path__` (maybe better name for heading)

fabb: i added a little paragraph for explaining the use case. makes sense?

fabb:
@evilebottnawi can we merge the PR?

evilebottnawi:
@fabb yep, at this week we merge and release, in my todo

prontiol:
I remember you was against it, @evilebottnawi?

Quotes are yours:
> Because it is fix only logic for file-loader, but other loader also can emit assets, i think we need solve this on webpack side, so i request minimum reproducible test repo

Does this PR fix this on webpack side? Where's the "minimum reproducible test repo"?

> We don't want add new option to file-loader, because in webpack@5 it was built-in loader, so we need solve this problem on webpack level to avoid same problems in future for other developers

You refused to merge my [PR](https://github.com/webpack-contrib/file-loader/pull/327) even if it was not adding any new options, but merged this one with an additional option?

evilebottnawi:
@prontiol A similar feature was implemented on webpack side and several people requested a similar solution for file-loader.

Sometimes I'm mistaken, I should have heard you better, sometimes it’s difficult because of the influx of many ideas, most of which look terrible, sorry

PR #369, opened 2020-03-17
chore(deps): update

evilebottnawi:
BREAKING CHANGE: use `md4` by default for hashing

This PR contains a:

- [x] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [x] **test update** 
- [ ] **typo fix**
- [x] **metadata update**

### Motivation / Use-Case

Use `md4` by default for perf reasons

### Breaking Changes

Yes

### Additional Info

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,14 +466,14 @@ Same as above, but with a customizable number of emojis
 ### `[hash]`
 
 Type: `String`
-Default: `md5`
+Default: `md4`
 
 Specifies the hash method to use for hashing the file content.
 
 ### `[contenthash]`
 
 Type: `String`
-Default: `md5`
+Default: `md4`
 
 Specifies the hash method to use for hashing the file content.
 
@@ -495,10 +495,9 @@ base49, base52, base58, base62, base64, and hex.
 #### `hashType`
 
 Type: `String`
-Default: `'md5'`
+Default: `'md4'`
 
-The type of hash that the has function should use. Valid values include: `md5`,
-`sha1`, `sha256`, and `sha512`.
+The type of hash that the has function should use. Valid values include: `md4`, `md5`, `sha1`, `sha256`, and `sha512`.
 
 #### `length`
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,9 +31,9 @@
       }
     },
     "@babel/compat-data": {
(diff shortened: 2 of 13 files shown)


jens-duttke:
May I ask why the default has been switched to MD4? Referring Wikipedia, the risk of a collision is much higher than for MD5 hashes.

I assume it can't be the speed on modern computers.

evilebottnawi:
@jens-duttke for perf reasons, webpack also uses md4 for hashing

> risk of a collision

No, not in out cases

mrlubos:
> @jens-duttke for perf reasons, webpack also uses md4 for hashing
> 
> > risk of a collision
> 
> No, not in out cases

Can you explain why please?

krysopath:
when you retire a hash algo for security reasons, then it would be also a pretty valid reason to switch to sha256 algo or similar. I cant see the reason at all.

odensc:
I'd assume what @evilebottnawi meant is that since the hash is used for caching purposes (not cryptographic purposes) it doesn't matter if there's a high collision risk, because the consequences in the event of a collision are close to nil.

However a non-cryptographic hash function such as xxHash would be much better suited to this use-case than an obsolete cryptographic hashing function, and would most likely be even faster.

It would be nice to get some insight into the reasoning behind this change from the maintainer themselves.

evilebottnawi:
> I'd assume what @evilebottnawi meant is that since the hash is used for caching purposes (not cryptographic purposes)

Yes

> However a non-cryptographic hash function such as xxHash would be much better suited to this use-case than an obsolete cryptographic hashing function, and would most likely be even faster.

yes, but we can't use it because it is require library compilation and can incompatibility with some platforms  

Why `md4`?

- not cryptographic purposes
- integration `xxHash` can be problematic
- `md4` and `md5` has collisions 
- `md4` is faster than `md5`

So using `md4` is best solution right now

tomaszs:
This change is marked as breaking change. How to adjust file-loader to comply with this change?

digilist:
@tomaszs I do not know the intention of the library authors for sure, but I guess they chose to bump the major version, because the build result will be different and so all caches will be invalidated. So I think there is no change necessary and you can simply update (this is what I did).

Issue #382, opened 2020-08-25
Pass additional assetInfo object when calling emitFile


bschlenk:
- Operating System: MacOS
- Node Version: 12.6.1
- NPM Version: 6.14.7
- webpack Version: 4.29.6
- file-loader Version: 3.0.1

### Feature Proposal

Webpack loader's [emitFile](https://github.com/webpack-contrib/file-loader/blob/master/src/index.js#L59) call takes a 4th param that can provide additional info about an asset, like whether it is immutable or not. `file-loader` should pass in `{ immutable: true }` when the file name includes `[contenthash]` so that other plugins can benefit from this info.

The new `assetInfo` param was added in https://github.com/webpack/webpack/issues/9038.

### Feature Use Case

I'm writing a plugin that uses this info to determine what cache-control headers a file can be given.

evilebottnawi:
Let's do it

evilebottnawi:
Fixed by https://github.com/webpack-contrib/file-loader/pull/383

PR #383, opened 2020-08-26
feat: pass immutable info on files

bschlenk:
This PR contains a:

- [ ] **bugfix**
- [x] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [ ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Passes an `immutable` flag through assetInfo which other plugins may use.

https://github.com/webpack-contrib/file-loader/issues/382

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,16 +14,14 @@ export default function loader(content) {
   });
 
   const context = options.context || this.rootContext;
+  const name = options.name || '[contenthash].[ext]';
+  const immutable = /\[([^:\]]+:)?(hash|contenthash)(:[^\]]+)?\]/gi.test(name);
 
-  const url = loaderUtils.interpolateName(
-    this,
-    options.name || '[contenthash].[ext]',
-    {
-      context,
-      content,
-      regExp: options.regExp,
-    }
-  );
+  const url = loaderUtils.interpolateName(this, name, {
+    context,
+    content,
+    regExp: options.regExp,
+  });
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -14,16 +14,14 @@ export default function loader(content) {
   });
 
   const context = options.context || this.rootContext;
+  const name = options.name || '[contenthash].[ext]';
+  const immutable = /\[([^:\]]+:)?(hash|contenthash)(:[^\]]+)?\]/gi.test(name);
```
bschlenk, line 18: Ideally this would be another function in https://github.com/webpack/loader-utils

evilebottnawi: Yes, I will test it deeply in near future

bschlenk:
not sure why [ModuleConcatenationPlugin test](https://github.com/webpack-contrib/file-loader/pull/383/checks?check_run_id=1031775282) is failing. [Source](https://github.com/webpack/webpack/blob/820174baef91f26839ee0cefa2cfb4c1785c3f68/lib/optimize/ModuleConcatenationPlugin.js) doesn't use the `immutable` flag, is it unrelated?

evilebottnawi:
I think we should improve test, webpack@5 implements better tree so count of modules can be reduced

evilebottnawi (approved):
Thanks

evilebottnawi:
I will merge and fix tests after merge

PR #387, opened 2020-10-02
fix syntax error in function example for name.

Taimoor0217:
This PR contains a:

- [ ] **bugfix**
- [ ] new **feature**
- [ ] **code refactor**
- [ ] **test update** 
- [x ] **typo fix**
- [ ] **metadata update**

### Motivation / Use-Case

Noticed a small typo in the [function example](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksname) of the name option.

### Breaking Changes

This is only a minor fix

### Additional Info
Screen shot attached.

file-loader-colon-error

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ module.exports = {
         test: /\.(png|jpe?g|gif)$/i,
         loader: 'file-loader',
         options: {
-          name(resourcePath, resourceQuery) {
+          name:(resourcePath, resourceQuery) {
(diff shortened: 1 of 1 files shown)


evilebottnawi (changes requested):
It is a method, please read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

